PR #3, opened 2015-01-05
Update sairouter.h

tushar-ty:
Fix io descriptor.


lguohan:
the comment is not correct, the vr_id in sai_create_virtual_router_fn should be out.

PR #9, opened 2015-01-24
RIF ID is more relevent for neighbour entry. Added RIF ID and removed the

AshokDaparthi:
...e

VR_ID for neighbour entry.


lguohan:
merged into master branch. thanks

PR #10, opened 2015-01-28
Update saitypes.h

tushar-ty:
Add native 32 bit data types in the sai_attribute_value_t union as a lot of attributes are of the same type in various modules.


lguohan:
added sai_uint32_t and sai_int32_t in the commit to make the header style more consistent.

PR #12, opened 2015-01-30
SAI LAG/Hash API Proposal - Base Version

arunsubash:
SAI LAG/Hash API Proposal for Review


lguohan:
merged to the main branch

PR #14, opened 2015-02-04
SAI defaults for each read write attribute are added. Notes for few API. Corrected the data types for attributes.

AshokDaparthi:


lguohan:
157  -    /* Operational Status [sai_port_oper_status_t] */ 
 157 +    /* Operational Status [sai_port_oper_status_t]  
 158 +       (default to SAI_PORT_OPER_STATUS_DOWN */ 
158 159      SAI_PORT_ATTR_OPER_STATUS, 

What does it mean by having a default value for this attribute? Should this value be always reflec the hardware status? If it is up in hw, then it is up here. If it is down in the hw, the it is down here.

sai/inc/saiport.h
```diff
@@ -154,7 +154,8 @@ typedef enum _sai_port_attr_t
     /* Port Type [sai_port_type_t] */
     SAI_PORT_ATTR_TYPE,
 
-    /* Operational Status [sai_port_oper_status_t] */
+    /* Operational Status [sai_port_oper_status_t] 
+       (default to SAI_PORT_OPER_STATUS_DOWN */
```
lguohan: What does it mean by having a default value for this attribute? Should this value be always reflect the hardware status? If it is up in hw, then it is up here. If it is down in the hw, the it is down here.

sai/inc/sairouter.h
```diff
@@ -67,15 +70,17 @@ typedef enum _sai_virtual_router_attr_t
...
 *    [out] vr_id - virtual router id
 *    [in] attr_count - number of attributes
 *    [in] attr_list - array of attributes
-*
+* 
 * Return Values:
 *    SAI_STATUS_SUCCESS on success
 *    Failure status code on error
+* Note: SAI_SWITCH_ATTR_SRC_MAC_ADDRESS | 
+* SAI_VIRTUAL_ROUTER_ATTR_SRC_MAC_ADDRESS should be set before create.
```
lguohan: Do you mean this?

Note:
  - SAI_SWITCH_ATTR_SRC_MAC_ADDRESS should be set if SAI_VIRTUAL_ROUTER_ATTR_SRC_MAC_ADDRESS is not passed in the create call. Otherwise, SAI_STATUS_ADDR_NOT_FOUND is returned.

sai/inc/saiport.h
arunsubash: Dell - Internal Use - Confidential
Depending on the NPU the default state in HW can be OPER_UP/DOWN. The SAI could put the port as ADMIN_DOWN by default when 1st created which could have same or different from the HW default state for a port depending on the NPU

-Arun

sai/inc/saitypes.h
```diff
@@ -233,6 +233,10 @@ typedef union {
     sai_acl_action_data_t acldata;
 } sai_attribute_value_t;
 
+/* 
+ * Data Type to use enum's as attribute value is sai_int32_t s32
+ *
```
lguohan: - This comment is better to moved to line 215 as it is related to attribute value.

sai/inc/sairouter.h
AshokDaparthi: Yes Guohan, MAC address is madatory to creating the Virtual interface.

lguohan:
merged into master branch.

PR #18, opened 2015-02-05
SAI buffer base version

itaibaz:


lguohan:
committed as one commit

PR #28, opened 2015-02-24
fix typo in next ho group + route prio comment

itaibaz:


sai/inc/sairoute.h
```diff
@@ -44,7 +44,7 @@ typedef enum _sai_route_attr_t
     SAI_ROUTE_ATTR_PACKET_ACTION,
 
     /* Packet priority for trap/log actions [uint8_t] 
-       (default to 0) */
```
lguohan: are we going to use this attribute eventually? I thought we are going to use the priority defined in the host group,

itaibaz: IMO, yes
This is used for configuring trap priority for a certain routing rule with action trap/mirror
Whereas host if property is used to configure the priority for trap group, and then we correlate the given set of traps to groups

arunsubash: What will be expected default or its a mandatory attribute

itaibaz: the attribute can’t be mandatory, as it’s not relevant/needed in case of action=drop or action=forward
even making it mandatory in case of action=trap or action=log doesn’t seem necessary
I think we will have to further expose the priorities each vendor has, following the discussion we had in our last face to face, about exposing HW resources.
And as default, we can use 0 = lowest priority.

arunsubash: Fine The confusion is because I saw partial diffs and though you were removing the comment that had the default value

lguohan: what's "trap priority for a certain routing rule"? As for my understanding, you set up a trap Id for certain routing rule, then you map the trap Id to a trap group. In the trap group, you can configure the priority which is related to the scheduling on the cpu port. What do you mean by "trap priority" here?

lguohan:
merged

PR #32, opened 2015-03-07
Added the SAI_API_MAX to track number of sai_api tables supported

atitjain:
and status for read only attributes.


MUBIRUSAMUEL:
I think the Added the SAI_API_MAX to track number of sai_api tables supported

PR #34, opened 2015-03-11
Datastruture for lit of integers

arunsubash:
The new data structure would be used for cases where the attribute needs to
fill in a list of integers. There are quite a few instances of that as we
define more objects. This would help to have one common data structure to
use rather than each object defining their own


sai/inc/saitypes.h
```diff
@@ -233,6 +233,11 @@ typedef struct _sai_acl_action_data_t
 
 } sai_acl_action_data_t;
 
+typedef struct _sai_gen_list_t {
```
lguohan: is it ok to rename it as sai_enum_list_t? gen_list sounds like a too generic name?

arunsubash: Yes changing a name is fine with me

lguohan:
checked in the change in another pull request

PR #38, opened 2015-03-24
SampleFlow+Mirror Update

arunsubash:
Renamed SFLOW as SampleFlow
Mirror removed L2 Tunnel header option for ERSPAN and made the ERSPAN related attributes as Create+Set


lguohan:
merged

PR #39, opened 2015-03-31
Proposed SAI ACL Changes 

ramangupta:
The change is to add support for PBR which is added as an ACL Rule Action Attribute and IPv6 Flow Label which is added as an ACL Rule Field Attribute.

Also boundaries to enum have been added for well defined range of ACL attributes like fields and actions.


sai/inc/saiacl.h
```diff
@@ -309,18 +318,29 @@ typedef enum _sai_acl_entry_attr_t
...
+    SAI_ACL_ENTRY_ATTR_ACTION_START = 0x00002000,
+
     /* Forward Normally */
-    SAI_ACL_ENTRY_ATTR_ACTION_FORWARD = 0x00002000,
+    SAI_ACL_ENTRY_ATTR_ACTION_FORWARD = SAI_ACL_ENTRY_ATTR_ACTION_START,
 
     /* Redirect Packet */
     SAI_ACL_ENTRY_ATTR_ACTION_REDIRECT,
 
+    /* Redirect to Nexthop */
```
lguohan: with uoid, the existing SAI_ACL_ENTRY_ATTR_ACTION_REDIRECT should take [sai_object_id_t] as the parameter, and it can be sai_port, sai_next_hop, sai_next_hop group, and etc. on. Therefore, the existing entry will cover this new case. 

How about add following comments for SAI_ACL_ENTRY_ATTR_ACTION_REDIRECT
/* Redirect Packet to a destination which can be a port, lag, nexthop, nexthopgroup. [sai_object_id_t] */

lguohan:
Hi Raman,
   I have commented this on the github. Please take a look at.

Thanks,

-Guohan

sai/inc/saiacl.h
```diff
@@ -569,7 +593,10 @@ typedef sai_status_t (*sai_get_acl_entry_attribute_fn)(
 *   [in] attr_list - array of attributes
 *
 * Return Values:
-*    SAI_STATUS_SUCCESS on success
+*    SAI_STATUS_SUCCESS on success (default = byte counter)
+*    SAI_STATUS_INVALID_PARAMETER - If both
```
lguohan: can we move such comments to the attributes? Since now all the create functions have same signature, we plan to consolidate all them. The idea to create a one create function prototype, like 

typedef sai_status_t (*sai_create_sai_object_fn)(_Out sai_object_id_t *object_id, _In_ attr_count, _In_ *attr_array);

and use this one to define all create functions that have same signature. The idea to focus on the comments for attributes, not the functions.

sai/inc/saiacl.h
```diff
@@ -309,16 +319,25 @@ typedef enum _sai_acl_entry_attr_t
     /* Ip Frag [sai_acl_ip_frag_t] */
     SAI_ACL_ENTRY_ATTR_FIELD_IP_FRAG,
 
+    /* IPv6 Flow Label */
```
lguohan: this needs to be replicated in SAI_ACL_TABLE_ATTR_FIELD as well.

lguohan:
merged the changes into v0.9.2. a few issues during the merge:
1. your patch is generated between opencomputeproject:master and ramangupta:v0.9.2. If you want to check code into opencomputerproject:v0.9.2, please base on opencomputerproject:v0.9.2.
2. This pull request history has quite a few abandoned code. Please consider consolidate your pull request history first. I had consolidate the history. As a result, the committer becomes to me.

PR #40, opened 2015-04-02
Remove SAI_NEXT_HOP_ATTR_TYPE from sainexthop.h.

jiacao:
SAI_NEXT_HOP_ATTR_TYPE is not needed any more, since there is only one
type of next hop, which is IP next hop.


sai/doc/SAI-Proposal-11-Tunnel.md
```diff
@@ -0,0 +1,488 @@
...
+
+### L3 Tunnel Egress Interface Model ###
+
+There are two types of L3 egress interface defined in the SAI, namely **router interface** and **l3 tunnel egress interface**. To note, the router interface object is used for both ingress and egress interface. As shown in Figure 1, the sai\_next\_hop object can point to either a router interface or a tunnel egress interface. A tunnel egress interface must point to a router interface. In this model, the tunnel egress interface is responsible for adding the tunneling header, e.g., IP, GRE, NVGRE. The router interface is responsible for updating the Ethernet source mac and vlan header. The next hop provides the destination mac information. When multiple tunnel egress interfaces point to a same router interface, all these tunnels will share a same source mac address and vlan id.
+
+As specified in earlier spec, a route object can point to either a next hop object or a next hop group object. In case the route points to a next hop group, and the next hop group object points to multiple next hops, and each next hop points to a tunnel egress interface. This will spread the IP packets over a set of L3 tunnels. 
+
```
SathishKumarTR-Dell: Actually, as per this proposal, each next hop pointing to a tunnel egress interface does not necessarily mean the packets are spanning over different L3 tunnels. It will be the case even for next hops of a single L3 tunnel having ECMP paths to reach the tunnel destination.
For eg, for a L3 tunnel having DIP (100.0.0.1) which is reachable by 3 next-hops (10.0.0.1/RIF1), (11.0.0.1/RIF2) and (20.0.0.1/RIF3) - user will have to create 3 tunnel egress interface objects (with same attributes except the ROUTER_INTF attribute) and then create the 3 tunnel next-hop objects.

Also, setting up tunnel termination for the same tunnel does not have any relation to the number of egress router interfaces for the tunnel destination. So creating tunnel termination on each tunnel egress interface object in the above example may not fit as well.

Will it be useful to have a single tunnel interface object and capability to associate it with multiple egress router interfaces?

sai/inc/sail3tunnelintf.h
```diff
@@ -0,0 +1,162 @@
...
+/*
+ *  Attribute id for l3 tunnel egress interface
+ */
+typedef enum _sai_l3_tunnel_egress_interface_attr_t
+{
+    /* READ-ONLY */
+
+    /* L3 tunnel egress interface router interface. [sai_object_id_t] (MANDATORY_ON_CREATE|CREATE_ONLY) */
+    SAI_L3_TUNNEL_EGRESS_INTERFACE_ATTR_ROUTER_INTF,
+
```
SathishKumarTR-Dell: Please refer to the note posted at line 27 in the SAI-Proposal-11_Tunnel.md above. Having an one-to-one association with tunnel interface and egress router interface means multiple tunnel interfaces needs to be created for tunnel over ECMP paths.

sai/doc/SAI-Proposal-11-Tunnel.md
atitjain: next hop group is for ECMP only. I think we will need to have multiple next-hops because each route interface may lead to different routers and hence might have different DMAC and may be vlan. Only common thing is SIP and DIP of the tunnel, the next hop attributes are different for each path.

sai/doc/SAI-Proposal-11-Tunnel.md
```diff
@@ -0,0 +1,488 @@
...
+(Current proposal only defines APIs for IP-in-IP tunnel egress interfaces, others are left for the future)
+
+### L3 Tunnel Egress Interface Model ###
+
+There are two types of L3 egress interface defined in the SAI, namely **router interface** and **l3 tunnel egress interface**. To note, the router interface object is used for both ingress and egress interface. As shown in Figure 1, the sai\_next\_hop object can point to either a router interface or a tunnel egress interface. A tunnel egress interface must point to a router interface. In this model, the tunnel egress interface is responsible for adding the tunneling header, e.g., IP, GRE, NVGRE. The router interface is responsible for updating the Ethernet source mac and vlan header. The next hop provides the destination mac information. When multiple tunnel egress interfaces point to a same router interface, all these tunnels will share a same source mac address and vlan id.
```
atitjain: "The router interface is responsible for updating the Ethernet source mac and vlan header. The next hop provides the destination mac information."
Above statement it not correct for current SAI model. Currently the route give next-hop and next-hop has next hop ip and router interface. These two are used as the key to the neighbor table which gives DMAC.

jiacao: Yes. The previous statement is not accurate. I changed it to: 
"The next hop provides the next hop IP address, which can be used to find out the next hop mac address from the neighbor ARP table by the switch"

sai/doc/SAI-Proposal-11-Tunnel.md
jiacao: The reason we choose the current design is because of its flexibility. If we bind multiple router interfaces to a single tunnel interface to implement ECMP group. There are several problems: 
1. Among the bound router interfaces, what if their next hop IP addresses are different? Say, interface A wants next hop 10.0.0.1, while interface B wants next hop 10.0.0.2. 
2. What if a ECMP contains both tunnel and router interfaces?

With our proposed design, we can implement the above scenarios like this: 
1. Create a next hop group, containing different next hops. Each next hop points to a tunnel interface. Each tunnel interface points to a router interface. 
2. Create a next hop group, containing different next hops. Some of the next hop point to a tunnel interface, some of them point a router interface.

sai/inc/sail3tunnelintf.h
jiacao: Correct. Multiple tunnel interfaces need to be created for the ECMP case.

jiacao:
Thanks for everyone's valuable comments. We'll cancel this pull request, and submit a new one later.

PR #44, opened 2015-04-08
Defining General List and defining object ID in attribute value

arunsubash:
Defined general uint32 and int32 list which can be used in attributes that needs a list of integers
Defined sai_object_t in attribute so all attributes using object ID can use this instead of u64


lguohan:
merged

PR #45, opened 2015-04-09
New Header files - LAG/Hash, STP, Mirror, Sample Packet

arunsubash:
New Header files - LAG/Hash, STP, Mirror, Sample Packet
Port Numbering Change
Compile fixes
Merge changes


sai/inc/saiport.h
```diff
@@ -135,15 +110,58 @@ typedef enum _sai_port_internal_loopback_mode_t
 */
 typedef enum _sai_port_fdb_learning_mode_t 
 {
+    /* Drop unknown Source MAC. Do not learn */
+    SAI_PORT_LEARN_MODE_DROP,
```
lguohan: what's the difference between drop mode and disable mode?

lguohan:
I have consolidated them into one commit and checked in to the v0.9.2 branch. It's unclear to me the difference between MODE_DROP and the MODE_DISABLE. Let's have a discussion on this before we check in this.

sai/inc/saiport.h
atitjain: One possibility is:
In disable mode there should be no learning and all the packets should be flooded. In the DROP mode user should be managing FDB completely and add all entries as static. If any entry is not added by user, drop the packet.

arunsubash: /*
* Attribute data for SAI_PORT_ATTR_FDB_LEARNING
*/
typedef enum _sai_port_fdb_learning_mode_t
{
    /* Drop unknown Source MAC. Do not learn. Do not forward*/
    SAI_PORT_LEARN_MODE_DROP,

     /* Do not learn unknown source MAC. Forwarding will occur based on destination MAC */
     SAI_PORT_LEARN_MODE_DISABLE,

     /* Hardware learning. Learn source MAC. Forward based on destination MAC*/
     SAI_PORT_LEARN_MODE_HW,

    /* Trap the unknown source MAC to CPU. Do not learn. Do not forward */
    SAI_PORT_LEARN_MODE_CPU,

} sai_port_fdb_learning_mode_t;

The comments had details of each mode

Drop - If there is no SRC SA match we will drop the packet
HW Learn Disable -  If there is no SRC SA match we will forward or flood based on dest MAC

sai/inc/sailag.h
```diff
@@ -0,0 +1,123 @@
...
+
+/*Lag attribute: List of attributes for LAG module*/
+typedef enum _sai_lag_attr_t {
+
+    /** READ_WRITE */
+
+    /* SAI port list [sai_object_list_t] (MAXDATORY_ON_CREATE|CREATE_AND_SET)
+     * The LAG must have at least one SAI port member at the creation time 
+     */
+    SAI_LAG_ATTR_PORT_LIST,
```
atitjain: I think this should be read only attribute. Internally gets updated based on the add/delete port. Max ports in LAG, if required could be there as an attribute.

sai/inc/saisamplepacket.h
```diff
@@ -0,0 +1,139 @@
...
+
+    /** READ-WRITE */
+
+    /** MANDATORY_ON_CREATE|CREATE_AND_SET */
+    /** Sampling rate of type uint32_t. The sampling rate specifies 
+      * random sampling probability as the ratio of packets observed 
+      * to samples generated. For example a sampling rate of 100 specifies 
+      * that, on average, 1 sample will be generated for every 100 packets 
+      * observed. */   
+    SAI_SAMPLEPACKET_ATTR_SAMPLE_RATE,
```
atitjain: We discussed about the SAI cache refresh rate. Are we planning to add it.

arunsubash: Yes it would be added as a switch attribute in the coming week

sai/inc/saimirror.h
```diff
@@ -0,0 +1,207 @@
...
+    /** tunnel source IP - [sai_ip_address_t] */
+    SAI_MIRROR_SESSION_ATTR_SRC_IP_ADDRESS,
+
+    /** MANDATORY_ON_CREATE|CREATE_AND_SET */
+    /** tunnel destination IP - [sai_ip_address_t] */
+    SAI_MIRROR_SESSION_ATTR_DST_IP_ADDRESS,
+
+    /** MANDATORY_ON_CREATE|CREATE_AND_SET */
+    /** L2 source MAC address - [sai_mac_t] */
+    SAI_MIRROR_SESSION_ATTR_SRC_MAC_ADDRESS,
```
atitjain: Do we need to configure SA/DA here or depending on SIP/DIP next-hop should be installed by control plane.

sai/inc/saimirror.h
```diff
@@ -0,0 +1,207 @@
...
+
+    /** MANDATORY_ON_CREATE|CREATE_AND_SET */
+    /** L2 destination MAC address - [sai_mac_t] */
+    SAI_MIRROR_SESSION_ATTR_DST_MAC_ADDRESS,
+
+    /** MANDATORY_ON_CREATE|CREATE_AND_SET */
+    /** GRE protocol Id - [uint16_t] */
+    SAI_MIRROR_SESSION_ATTR_GRE_PROTOCOL_TYPE,
```
atitjain: Not sure but this is the ethertype of the payload. I think this should be copied from mirror packet to the GRE header in ideal case. Or we can mirror only certain packet types.

sai/inc/sailag.h
arunsubash: This is an additional way to set all the LAG members in one shot during create and there after used to query the list of LAG members

sai/inc/saimirror.h
arunsubash: The SA/DA values are programmed by the control plane. The <Src IP, Dest IP, SA, DA> All 4 are controlled by control plane and they program them based on config and reachability. They get reprogrammed if any of them changes due to config or reachability change

sai/inc/saimirror.h
arunsubash: Its not copied from the inner packet. Currently it has to be set to the ERSPAN protocol type

atitjain: Thanks for clarification.

sai/inc/sailag.h
atitjain: Thanks for clarification.

sai/inc/saimirror.h
lguohan: auto id can be added in the future if needed.

sai/inc/saiport.h
lguohan: @atitjain , in disable mode whether packets are flooded, dropped, or unicast depends on whether the destination mac is known or not. If it is know, it is unicast. If it is unknown, it can be flooded or dropped based on SAI_SWITCH_ATTR_FDB_UNICAST_MISS_ACTION.

@arunsubash : added to v0.9.2, please check.

sai/inc/sailag.h
lguohan: @arunsubash : are you suggesting this should be MANDANTORY_ON_CREATE and CREATE_ONLY since it cannot be set after created?

sai/inc/saimirror.h
lguohan: @atitjain : are you suggesting the asic is going to do a l3 lookup based on mirror destination ip?

sai/inc/saiport.h
atitjain: My understanding is, that this is learning disabled so the FDB will not learn anything. Ok if still static entries could be added then your interpretation is correct.

sai/inc/saimirror.h
atitjain: Given its GRE tunnel origination, I was hoping this. So based on the next hop for the GRE tunnels DIP the arp should resolve and nh should be installed.

sai/inc/saiport.h
lguohan: the mental model here is that l2 forwarding has two stages, one is learning for the src mac, the second is forwarding based on the dest mac. This sai_port_fdb_learning_mode_t controls the first stage behavior, 1) whether to learn the mac or not, 2) for unknown mac, whether to proceed to the second stage or not.

Then the second stage controls the behavior whether the dest mac can be found or not. We'd like to separate these two stages. 

Does this model make sense to you?

sai/inc/saimirror.h
lguohan: as you known, there are certain asics do not follow the pipeline model you describe above. I also believe your pipeline model makes sense although I do not know if there is any asic that has implemented that or not since the model you describe require recirculate the mirrored packets to l3 lookup.

The question now becomes how does sai deal with these two models. does it provides two models or one model? In case we want to go with one model, it will be similar to what we did for the neighbor table. When the asic does not have neighbor table, the software maintains a copy for lookup. As to this case, we might need to maintain the whole l3 route table, as well as nexthop group, next hop table. Lots of states will be maintained in the software now.

atitjain: Understand your point about re circulation. Agree that it may not be possible in pipelines to do these multiple look-ups etc in one pass. Lets retain these attributes.

sai/inc/saiport.h
atitjain: Completely agree to this model with one additional point that these 2 stages, SA learning and DA look-up have correlation. Let me list the attributes and my interpretation:

SAI_PORT_LEARN_MODE_DROP - says if SA is not hit in the FDB table then don't learn it and dont forward frame. So MAC is never learnt dynamically. If a MAC matching packets SA is added statically in FDB table then it would not be missed in SA look-up stage, further packet will be forwarded based on DA look-up and 
SAI_SWITCH_ATTR_FDB_UNICAST_MISS_ACTION

SAI_PORT_LEARN_MODE_DISABLE - The MAC will never be learnt dynamically because learning is disabled. But packet will be forwarded based on DA hit and SAI_SWITCH_ATTR_FDB_UNICAST_MISS_ACTION. DA hit in this case can happen only if the a MAC matching packets DA was added as static entry.

Let me know if this is correct.

sai/inc/sailag.h
arunsubash: Yes Guohan this is most likely to be used 1st time creating the LAG when we specific the list of ports. There after we may be just adding and removing ports and its unlikely to set the entire set of ports again. Anyway if a user wants to set the entire new list of ports after addition/deletion in one shot this can be used. I am OK to remote the CREATE_ONLY so it can be used after that as well

atitjain: Should we impost a MANDATORY_ON_CREATE. User can always define the LAG and its attributes and add ports later. I understand unless a port is added the LAG is not usable, but why to impose restriction. 
When a port in the LAG goes down, what is the mechanism to indicate LAG. Is it sai_remove_port_from_lag_fn? Similarly when it comes back again?

arunsubash: I am fine to remove both MANDATORY_ON_CREATE and CREATE_ONLY and make it like any other read/write attribute to give maximum flexibility
Yes Port UP/DOWN handling is done by application and its expected to do a LAG port add/remove based link status
@lguohan please confirm i will make the changes

lguohan: I am fine with remove MANDATORY_ON_CREATE and keep CREATE_AND_SET. Looks like vendor SDK do support creating LAG without any member.

arunsubash: @lguohan Yes will make this change

PR #46, opened 2015-04-10
Merging the latest API proposals to keep it upto date with the headers

arunsubash:
Checking in the latest proposals for LAG, STP, samplepacket, Mirror, Port Numbering, Host Interface. This would make it on par with the checked in headers except for Host Interface which is not checked in yet


lguohan:
merged

PR #47, opened 2015-04-10
Update FDB Learn Mode Comments + Counter refresh Interval

arunsubash:
Updated the comments for FDB Learn Modes
Add SAI_PORT_LEARN_MODE_DROP
Add switch attribute for counter refresh interval


sai/inc/saiswitch.h
```diff
@@ -242,6 +242,20 @@ typedef enum _sai_switch_attr_t
...
     SAI_SWITCH_ATTR_ECMP_MAX_PATHS,
 
+    /* This setting can be used to configure the SW counter cache refresh 
+     * rate. SW counter cache is used if the retrieval of counters 
+     * directly from HW for each counter read is CPU intensive.
+     *
+     * SW counter cache used      | Direct read from HW used
+     * -----------------------------------------------------
+     * Default - 1 sec            |  Default - 0 sec
+     * Set Range - (1 - 300) sec  |  Set not allowed
```
lguohan: what does this "Set not allowed" means? My understanding of this attribute is 
if it is 0, then it means direct read from HW. If it is larger than 0, it means read form SW counter cache, and the value is cache update interval. The maximum cache update interval is 300 seconds? Why do we have this restriction?

Get will always return the value.

arunsubash: My understanding is that the application will not have an option to tell SAI if it should use SW cache based approach or direct HW read. The reason is it depends on SDK implementation etc. Hence the set is allowed only allowed for SW based cache is enabled
300sec need not be max value it can be higher or we just don't define the range and allow user to configure max

lguohan: see if following comment is more clear to you?

This setting can be used to configure the counter read mode and the SW counter cache refresh rate. SW counter cache is used if the retrieval of counters directly from HW for each counter read is CPU intensive. 

0: Direct HW read approach 
>0: Software cache read approach. The value is the software cache update interval. Default update interval is 1 second.

SDK implementation may support both approaches or only one approach. In case only one approach is supported, switching from one approach to another will return SAI_STATUS_INVALID_ATTR_VALUE.

sai/inc/saiport.h
```diff
@@ -110,15 +110,20 @@ typedef enum _sai_port_internal_loopback_mode_t
...
-    /* Hardware learning */
+    /* Hardware learning. Learn source MAC. Forward based on destination MAC */
     SAI_PORT_LEARN_MODE_HW,
 
-    /* Trap the unknown source MAC to CPU. Do not learn */
+    /* Trap the unknown source MAC to CPU. Do not learn. Do not forward */
     SAI_PORT_LEARN_MODE_CPU,
 
+    /* Drop unknown Source MAC. Do not learn. Do not forward*/
+    SAI_PORT_LEARN_MODE_DROP,
```
lguohan: already in the v0.9.2 branch. please sync.

sai/inc/saiswitch.h
atitjain: @lguohan we discussed it as described by @arunsubash  on our mailing list. This is a SDK/SAI implementation dependent functionality. In case the SDK has caching of counters, this attribute is expose to the adapter host to decide its optimal nearest SFLOW polling interval.

sai/inc/saiport.h
arunsubash: Yes our changes overlapped. I will raise a new request after merging the changes

lguohan:
merged

PR #48, opened 2015-04-10
add api to query sai object type

lguohan:


arunsubash:
Looks OK to me

sai/inc/saitypes.h
```diff
@@ -132,6 +132,26 @@ typedef struct _sai_object_list_t {
...
+    SAI_OBJECT_TYPE_NEXT_HOP         =  4,
+    SAI_OBJECT_TYPE_NEXT_HOP_GROUP   =  5,
+    SAI_OBJECT_TYPE_ROUTER_INTERFACE =  6,
+    SAI_OBJECT_TYPE_ACL_TABLE        =  7,
+    SAI_OBJECT_TYPE_ACL_ENTRY        =  8,
+    SAI_OBJECT_TYPE_ACL_COUNTER      =  9,
+    SAI_OBJECT_TYPE_HOST_INTERFACE   = 10,
+    SAI_OBJECT_TYPE_MIRROR           = 11,
+    SAI_OBJECT_TYPE_SAMPLEPACKET     = 12,
+    SAI_OBJECT_TYPE_STP_INSTANCE     = 13,
```
SathishKumarTR-Dell: Will it be useful to define a SAI_OBJECT_TYPE_END here? That would help to return SAI_OBJECT_TYPE_NULL if object type does not fall in the range (>= SAI_OBJECT_TYPE_PORT && < SAI_OBJECT_TYPE_END) in the sai_object_type_query() API as defined above.

lguohan: added SAI_OBJECT_TYPE_MAX

arunsubash: @lguohan Previously we had agreed the object id would be the same for Port and LAG. In that case should we have seperate object types? We have LAG as one of the port type other than Logical and CPU

arunsubash:
@lguohan Previously we had agreed the object id would be the same for Port and LAG. In that case should we have seperate object types? We have LAG as one of the port type other than Logical and CPU

lguohan:
we have sai_create_lag(), from SAI perspective LAG is a sai object. What's the benefit for using a same type for both Port and LAG?

arunsubash:
We will remove PORT_TYPE_LAG so that these objects are independent and not part of the port numbering range

lguohan:
merged

PR #50, opened 2015-04-15
FDB Changes 

atitjain:
1. SAI_SWITCH_ATTR_FDB_BROADCAST_MISS_ACTION, is not possible. Its not there in the table
2. Added LIMIT reached and MOVE event types.
3. Added the limit voilation action in port and vlan


sai/inc/saiswitch.h
```diff
@@ -169,6 +169,11 @@ typedef enum _sai_switch_attr_t
      * zero means learning limit disable. (default to zero) */
     SAI_SWITCH_ATTR_MAX_LEARNED_ADDRESSES,
 
+    /* Action for packets with unknown source mac address
+     * when FDB learning limit is reached.
+     * [sai_packet_action_t] (default to SAI_PACKET_ACTION_DROP) */
+    SAI_SWITCH_ATTR_FDB_LEARNING_LIMIT_VIOLATION,
```
lguohan: I agree these need to be added. It unclear to me whether we should as them here or as host interface pipeline exception trap. If we add them there, we can point the trap to a trap group which defines the cpu queue and policer to do qos and rate-limiting.

PR #51, opened 2015-04-15
Update saitypes.h

tushar-ty:
Simplifying use of ACL field and action data types in the code.

### DIFF
--- a/sai/inc/saitypes.h
+++ b/sai/inc/saitypes.h
@@ -208,13 +208,26 @@ typedef struct _sai_acl_field_data_t
     /*
      * Field match mask
      */
-    uint64_t match_mask[2];
+    union {
+        sai_uint8_t u8;
+        sai_uint16_t u16;
+        sai_uint32_t u32;
+        sai_mac_t mac;
+        sai_ip4_t ip4;
+        sai_ip6_t ip6;
+    } match_mask;
 
     /*
      * Expected AND result using match mask above with packet field value.
      */
-    uint64_t match_data[2];
-
+    union {
+        sai_uint8_t u8;
+        sai_uint16_t u16;
+        sai_uint32_t u32;
+        sai_mac_t mac;
+        sai_ip4_t ip4;
+        sai_ip6_t ip6;
+    } match_data;
 } sai_acl_field_data_t;
 
 /*
@@ -230,8 +243,14 @@ typedef struct _sai_acl_action_data_t
     /*
      * Action parameter
      */
-    uint64_t parameter[2];
-
+    union {
+      sai_uint8_t u8;
+      sai_uint16_t u16;
+      sai_uint32_t u32;
+      sai_mac_t mac;
+      sai_ip4_t ip4;
+      sai_ip6_t ip6;
+    } parameter;
 } sai_acl_action_data_t;
 
 /*
--- a/sai/inc/saitypes.h
+++ b/sai/inc/saitypes.h
@@ -201,10 +201,9 @@ typedef enum _sai_acl_match_mode_t
 typedef struct _sai_acl_field_data_t
 {
     /*
-     * Field match mode
-     */
-    sai_acl_match_mode_t mode;
-
+     * match enable/disable
+     */ 
+    bool enable;
     /*
      * Field match mask
      */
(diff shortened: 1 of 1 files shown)


sai/inc/saitypes.h
```diff
@@ -208,13 +208,26 @@ typedef struct _sai_acl_field_data_t
...
      */
-    uint64_t match_mask[2];
+    union {
+        sai_uint8_t u8;
+        sai_uint16_t u16;
+        sai_uint32_t u32;
+        sai_mac_t mac;
+        sai_ip4_t ip4;
+        sai_ip6_t ip6;
+    } match_mask;
```
lguohan: change "match_mask" to "mask", also change "match_data" to "data"?

sai/inc/saitypes.h
```diff
@@ -208,13 +208,26 @@ typedef struct _sai_acl_field_data_t
...
-    uint64_t match_data[2];
-
+    union {
+        sai_uint8_t u8;
+        sai_uint16_t u16;
+        sai_uint32_t u32;
+        sai_mac_t mac;
+        sai_ip4_t ip4;
+        sai_ip6_t ip6;
+    } match_data;
```
lguohan: add "sai_object_list_t objlist" to the union for SAI_ACL_ENTRY_ATTR_FIELD_IN_PORTS.

add "sai_object_id_t oid" to the union for SAI_ACL_ENTRY_ATTR_FIELD_IN_PORT

sai/inc/saitypes.h
lguohan: I do not know if we really need the sai_acl_match_mode_t mode because for every match field, the mode is predetermined. I think the bool for enable/disable is ok for us.

sai/inc/saitypes.h
```diff
@@ -230,8 +243,14 @@ typedef struct _sai_acl_action_data_t
...
      */
-    uint64_t parameter[2];
-
+    union {
+      sai_uint8_t u8;
+      sai_uint16_t u16;
+      sai_uint32_t u32;
+      sai_mac_t mac;
+      sai_ip4_t ip4;
+      sai_ip6_t ip6;
```
lguohan: add  sai_object_id_t oid to the union for the attribute that needs oid?

tushar-ty:
Created pull request #55 with all the changes.

sai/inc/saitypes.h
ramangupta: Do we really need to replicate the union for field data/mask and action data. Can't we have a single union separately defined for all these data types and include them in field_data as data and mask whereas in action_data as simply data or parameter. 
This union which would contain all the attribute data types could further replace all the fields defined in sai_attribute_value_t . For instance
 
typedef union {
    bool booldata;
    char chardata[32];
    sai_uint8_t u8;
    sai_int8_t s8;
    sai_uint16_t u16;
    sai_int16_t s16;
    sai_uint32_t u32;
    sai_int32_t s32;
    sai_uint64_t u64;
    sai_int64_t s64;
    sai_port_id_t portid;
    sai_mac_t mac;
    sai_ip4_t ip4;
    sai_ip6_t ip6;
    sai_ip_address_t ipaddr;
    sai_port_list_t portlist;
    sai_next_hop_list_t nhlist;
} attribute_value;

typedef struct _sai_acl_field_data_t
{
    bool status;
    attribute_value data;
    attribute_value mask;
}

typedef struct _sai_acl_action_data_t
{
    bool status;
    attribute_value parameter;
} sai_acl_action_data_t;

and finally 

typedef union {
    attribute_value attribute_data; 
    sai_acl_field_data_t aclfield;
    sai_acl_action_data_t aclaction;
} sai_attribute_value_t;

lguohan:
this will make lots of existing code that use attribute_value break, and also make user use attribute_value slightly more difficult.

ramangupta:
In that case, can we leave sai_attribute_value_t as it is and change sai_acl_field_data_t and sai_acl_action_data_t as stated above. This would ensure all data types which could be used as ACL qualifiers or actions are visible to ACL field/action structures directly and no exceptions such as port list exist. 
ACL anyway needs to go through code churn to incorporate the changes.

tushar-ty:
The last proposal I made was based upon all the current ACL attribute requirements. If you could sight an example of what can't be accomplished based upon this version that will help, else whats been suggested looks like an overkill. But thats just my opinion, if this version does not fit any specific requirements then please create a new patch/pull request.

PR #52, opened 2015-04-16
add comments for acl match and action attributes

lguohan:
add comments for acl match and action attributes
consolidate packet action attributes (DROP/TRAP/LOG/FORWARD)
change COS to TC


sai/inc/saitypes.h
```diff
@@ -202,29 +202,15 @@ typedef struct _sai_ip_prefix_t {
...
  */
 typedef struct _sai_acl_field_data_t
 {
     /*
-     * Field match mode
-     */
-    sai_acl_match_mode_t mode;
+     * match enable/disable
+     */ 
+    bool enable;
```
lguohan: I do not know if we really need the sai_acl_match_mode_t mode because for every match field, the mode is predetermined. I think the bool for enable/disable is ok for us.

atitjain:
I have one quick question, shouldn't tables be created at the init with an attribute list and later should not allowed to change? To me it appears that for a given table we define the fields that could be selected to make keys in that table. Given that there are different TCAM considerations such as width of the TCAM etc, and these kind of things could not be changed dynamically.

atitjain:
Another question is do we want to keep number of entries per table dynamic? I think that would be difficult to manage with the priorities etc.

atitjain:
Another thing is I think we should have counter per entry rather than per table.

lguohan:
Current mode is that you specify matching fields for the acl table at creation. You cannot change that after creation. As indicated by this comment "(CREATE_ONLY, match fields cannot be changed after the table is created)"

Can you explain why dynamic table would be difficult to manage with the priorities? I do not understand the issue there.

I think we have counter per entry. see SAI_ACL_ENTRY_ATTR_ACTION_COUNTER

atitjain:
Thanks a ton Gouhan, I missed theses somehow. Thanks for the filling the GAPs

atitjain:
About the dynamic number of entries in the Table, The issue is that there would be some TCAM slices, tites etc and the table key selections could be done per slice/tile etc. Also some time multiple Tables could be mapped to the same tile/slice by selecting a superset of keys, if the max key width etc could be managed. Knowing The partition beforehand is required in such cases. Typically users will know the partition before hand. If this awareness is passed to SDK the TCAM could be managed better.

lguohan:
Let me try to understand what you are saying first. You are saying instead of asking the SDK to TCAM optimization, e.g. merging two table by combining their matching keys, it's better for the user application to do the optimization since it knows the requirement better. For example, user know what kind of acls he want to install, ip acl, mac acl, ipv6 acl, and etc. on. So, it's better for the user to carve out TCAM spaces instead of asking sdk/sai to do the optimization such as merge ip and mac acl together.

If my understanding is correct, then how does this related the dynamic table size. It seems to me that you are asking the SDK/SAI to dedicate the tcam slices to a table? Am I correct?

atitjain:
I am trying to fit the SAI model with some schemes of TCAM that I am aware of. So in these cases:
1. There are 'N' slices of TCAM with a given width say 'W' and depth 'D'
2. N, D and W could be configured for example I can make W" = 2W but then my D" = D/2
3. For each of the slice

PR #58, opened 2015-04-19
Host interface

mattyk:
update host interface header file :
the following objects has been added
trap_group,trap-id,user-defined trap-id
two additional user channel has been added:
file-descriptor,callback


sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+
+    SAI_HOSTIF_TRAP_ID_IGMP_TYPE_LEAVE = 0x00000007,
+
+    SAI_HOSTIF_TRAP_ID_IGMP_TYPE_V1_REPORT = 0x00000008,
+
+    SAI_HOSTIF_TRAP_ID_IGMP_TYPE_V2_REPORT = 0x00000009,
+
+    SAI_HOSTIF_TRAP_ID_IGMP_TYPE_V3_REPORT = 0x000000010,
+
+    SAI_HOSTIF_TRAP_ID_PACKET_SAMPLING = 0x00000011,
```
lguohan: this is hex, so should be 0x0000000b

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+
+    SAI_HOSTIF_TRAP_ID_BGP = 0x00002006,
+
+    SAI_HOSTIF_TRAP_ID_DHCPV6 = 0x00002007,
+
+    SAI_HOSTIF_TRAP_ID_OSPFV6 = 0x00002008,
+
+    SAI_HOSTIF_TRAP_ID_VRRPV6 = 0x00002009,
+
+    SAI_HOSTIF_TRAP_ID_BGPV6 = 0x00002010,
```
lguohan: 0x0000200a

sai/inc/saihostintf.h
```diff
@@ -125,22 +567,130 @@ typedef sai_status_t (*sai_set_host_interface_attribute_fn)(
...
+    * Can be either Port or LAG (MANDATORY_ON_SEND) */
+    SAI_HOST_INTERFACE_PACKET_EGRESS_PORT_LAG,
+
+    /* Egress LAG [sai_object_id_t] */
+    SAI_HOST_INTERFACE_PACKET_EGRESS_LAG,
+
+    /* Egress Queue [sai_object_id_t]. Optional */
+    SAI_HOST_INTERFACE_PACKET_EGRESS_QUEUE,
+
+    /* TX type [sai_tx_type_t]. (MANDATORY_ON_SEND) */
```
lguohan: should be sai_hostif_tx_type_t

sai/inc/saihostintf.h
```diff
@@ -125,22 +567,130 @@ typedef sai_status_t (*sai_set_host_interface_attribute_fn)(
...
+    sai_create_hostif_group_fn                     create_hostif_group;
+    sai_remove_hostif_group_fn   	               remove_hostif_group;
+    sai_set_hostif_group_attribute_fn              set_group_attribute;
+    sai_get_hostif_group_attribute_fn              get_group_attribute;
+    sai_create_hostif_trap_fn                      create_hostif_trap;
+    sai_remove_hostif_trap_fn   	               remove_hostif_trap;
+    sai_set_hostif_trap_attribute_fn               set_trap_attribute;
+    sai_get_hostif_trap_attribute_fn               get_trap_attribute;
+    sai_create_hostif_user_defined_trap_fn         create_hostif_user_defined_trap;
+    sai_remove_hostif_user_defined_trap_fn   	remove_hostif_user_defined_trap;
```
lguohan: please make the indentation consistent.

sai/inc/saihostintf.h
```diff
@@ -125,22 +567,130 @@ typedef sai_status_t (*sai_set_host_interface_attribute_fn)(
...
+
+    /* Ingress LAG [sai_object_id_t] (for receive-only) */
+    SAI_HOST_INTERFACE_PACKET_INGRESS_LAG,
+
+    /* Egress port [sai_object_id_t] (for send-only).
+    * Can be either Port or LAG (MANDATORY_ON_SEND) */
+    SAI_HOST_INTERFACE_PACKET_EGRESS_PORT_LAG,
+
+    /* Egress LAG [sai_object_id_t] */
+    SAI_HOST_INTERFACE_PACKET_EGRESS_LAG,
```
lguohan: why need this? you have EGRESS_PORT_LAG defined.

sai/inc/saihostintf.h
```diff
@@ -125,22 +567,130 @@ typedef sai_status_t (*sai_set_host_interface_attribute_fn)(
...
-    sai_set_host_interface_attribute_fn   set_host_interface_attribute;
-    sai_get_host_interface_attribute_fn   get_host_interface_attribute;
-
+    sai_create_host_interface_fn                   create_host_interface;
+    sai_remove_host_interface_fn                   remove_host_interface;
+    sai_set_host_interface_attribute_fn            set_host_interface_attribute;
+    sai_get_host_interface_attribute_fn            get_host_interface_attribute;
+    sai_create_hostif_group_fn                     create_hostif_group;
+    sai_remove_hostif_group_fn   	               remove_hostif_group;
+    sai_set_hostif_group_attribute_fn              set_group_attribute;
```
lguohan: set_hostif_group_attribute is more consistent with previous naming convention,

sai/inc/saihostintf.h
```diff
@@ -125,22 +567,130 @@ typedef sai_status_t (*sai_set_host_interface_attribute_fn)(
...
+    sai_get_hostif_group_attribute_fn              get_group_attribute;
+    sai_create_hostif_trap_fn                      create_hostif_trap;
+    sai_remove_hostif_trap_fn   	               remove_hostif_trap;
+    sai_set_hostif_trap_attribute_fn               set_trap_attribute;
```
lguohan: set_hostif_trap_attribute

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+    SAI_HOSTIF_TRAP_ID_CUSTOM_RANGE_BASE = 0x10000000,
+
+} sai_hostif_user_defined_trap_id_t;
+
+typedef enum _sai_hostif_user_defined_trap_attr_t
+{
+    /* Admin Mode enable/disable trap [bool] (default to TRUE) */
+    SAI_HOSTIF_USER_DEFINED_TRAP_ATTR_ADMIN_STATE,
+
+    /* trap channel to use [sai_hostif_trap_channel_t] (MANDATORY_ON_CREATE) */
```
lguohan: is this CREATE_AND_SET or CREATE_ONLY?

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+    SAI_HOSTIF_USER_DEFINED_TRAP_ATTR_TRAP_CHANNEL,
+
+    /* file descriptor [sai_object_id_t]
+    * Valid only when SAI_HOSTIF_TRAP_ATTR_TRAP_CHANNEL == SAI_HOSTIF_TRAP_CHANNEL_FD
+    *   (MANDATORY_ON_CREATE when SAI_HOSTIF_TRAP_ATTR_TRAP_CHANNEL ==
+    *    SAI_HOSTIF_TRAP_CHANNEL_FD)
+    */
+    SAI_HOSTIF_USER_DEFINED_TRAP_ATTR_FD
```
lguohan: can you change the FD after creation?

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+typedef enum _sai_hostif_group_attr_t
+{
+    /* Admin Mode [bool] (default to TRUE) */
+    SAI_HOSTIF_GROUP_ATTR_ADMIN_STATE,
+
+    /* group priority [uint32_t] (MANDATORY_ON_CREATE|CREATE_ONLY).
+    * This is equivalent to ACL table priority SAI_ACL_TABLE_ATTR_PRIORITY */
+    SAI_HOSTIF_GROUP_ATTR_PRIO,
+
+    /* cpu egress queue [uint32_t] (MANDATORY_ON_CREATE|CREATE_SET) */
```
lguohan: in the qos proposal, it will have this sai_queue_id_t, should we use it here?

lguohan:
throughout the header file, I see indentation is not consistent across the file, can you enforce tabstop=4 and expandtab?

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+
+    SAI_HOSTIF_TRAP_CHANNEL_CUSTOM_RANGE_BASE = 0x10000000
+} sai_hostif_trap_channel_t;
+
+typedef enum _sai_hostif_trap_attr_t
+{
+    /* Admin Mode enable/disable trap [bool] (default to TRUE) */
+    SAI_HOSTIF_TRAP_ATTR_ADMIN_STATE,
+
+    /* trap channel to use [sai_hostif_trap_channel_t] (MANDATORY_ON_CREATE) */
```
lguohan: this needs to be CREATE_AND_SET too, since we have the need to change the channel for default trap id.

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+*  [in] hostif_trap_id - host interface trap id
+*  [in] attr_count - number of attributes
+*  [in] attr_list - array of attributes
+*
+* Return Values:
+*    SAI_STATUS_SUCCESS on success
+*    Failure status code on error
+*/
+typedef sai_status_t(*sai_create_hostif_trap_fn)(
+    _In_ sai_hostif_trap_id_t hostif_trapid,
```
lguohan: I am thinking whether we can make trap_id as attribute of trap object, i.e., SAI_HOST_IF_TRAP_ATTR_MATCH_ID. Then, we can unified sai_create_object function signature here. 

as for the default trap object, we can have a switch attribute SAI_SWITCH_ATTR_DEFAULT_TRAP [sai_object_id_t]. After switch initialized, we'll create this default trap object. Then, we can set its trap object attribute, e.g., trap channel, trap group, but we'll not be able to delete it.

sai/inc/saihostintf.h
```diff
@@ -125,22 +567,130 @@ typedef sai_status_t (*sai_set_host_interface_attribute_fn)(
...
+typedef enum _sai_hostif_packet_attr
+{
+    /* Trap ID [sai_hostif_trap_id_t] (for receive-only) */
+    SAI_HOST_INTERFACE_PACKET_TRAP_ID,
+
+    /* Ingress port [sai_object_id_t] (for receive-only) */
+    SAI_HOST_INTERFACE_PACKET_INGRESS_PORT,
+
+    /* Ingress LAG [sai_object_id_t] (for receive-only) */
+    SAI_HOST_INTERFACE_PACKET_INGRESS_LAG,
```
atitjain: why mention LAG, the port unified object ID should be sufficient to indicate the type.

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
 
 #define HOST_INTERFACE_NAME_SIZE    16
 
+typedef enum _sai_hostif_group_attr_t
```
atitjain: This terminology looks confusing to me. It sounds like we are referring the hostif group and not the trap group. I think channel was a better term because we are carving various channels from a given interface based on the trap id. 

I understand we have used the channel for the 'method to trap' (CB, netdev, FD), please consider a more intuitive way. May be call it method or somethin.

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+typedef enum _sai_hostif_trap_channel_t
+{
+    /* receive packets via file desriptor  */
+    SAI_HOSTIF_TRAP_CHANNEL_FD,
+
+    /* receive packets via callback */
+    SAI_HOSTIF_TRAP_CHANNEL_CB,
+
+    /* receive packets via OS net device */
+    SAI_HOSTIF_TRAP_CHANNEL_NETDEV,
```
atitjain: By using these SAI_HOSTIF_TRAP_CHANNEL_FD and SAI_HOSTIF_TRAP_CHANNEL_NETDEV arent we binding things to some specific OS. I think this should be left to user. I understand some control plane will get ported easily using this, but why not leave it adapter host.

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+* Arguments:
+*  [in] hostif_user_defined_trap_id - host interface user defined trap id
+*  [in] attr_count - number of attributes
+*  [in] attr_list - array of attributes
+*
+* Return Values:
+*    SAI_STATUS_SUCCESS on success
+*    Failure status code on error
+*/
+typedef sai_status_t(*sai_create_hostif_user_defined_trap_fn)(
```
atitjain: Given that the SAI defined and user defined traps are separate spaces, we could handle them in one call: sai_create_hostif_trap_fn, the implementation can have its demultiplexing and then call internal functions to create either. Will keep things cleaner.

sai/inc/saihostintf.h
lguohan: ingress_port is enough here as you have the physical port to lag mapping at the control plane. agree with atit that ingress_lag can be removed

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+	/* fdb traps */
+	SAI_HOSTIF_USER_DEFINED_TRAP_ID_FDB_MIN = SAI_HOSTIF_USER_DEFINED_TRAP_ID_NEIGH_MAX + 1,
+
+	SAI_HOSTIF_USER_DEFINED_TRAP_ID_FDB_MAX = SAI_HOSTIF_USER_DEFINED_TRAP_ID_FDB_MIN + SAI_HOSTIF_USER_DEFINED_TRAP_ID_TABLE_RANGE,
+
+    SAI_HOSTIF_TRAP_ID_CUSTOM_RANGE_BASE = 0x10000000,
+
+} sai_hostif_user_defined_trap_id_t;
+
+typedef enum _sai_hostif_user_defined_trap_attr_t
```
atitjain: Why attributes here are different from _sai_hostif_trap_attr_t. What is the difference.

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+
+    /* trap channel to use [sai_hostif_trap_channel_t] (MANDATORY_ON_CREATE) */
+    SAI_HOSTIF_TRAP_ATTR_TRAP_CHANNEL,
+
+    /* file descriptor [sai_object_id_t]
+    * Valid only when SAI_HOSTIF_TRAP_ATTR_TRAP_CHANNEL == SAI_HOSTIF_TRAP_CHANNEL_FD
+    *   (MANDATORY_ON_CREATE when SAI_HOSTIF_TRAP_ATTR_TRAP_CHANNEL ==
+    *    SAI_HOSTIF_TRAP_CHANNEL_FD)
+    */
+    SAI_HOSTIF_TRAP_ATTR_FD,
```
atitjain: Do we have fd per trap. What about the netdevice then.

sai/inc/saihostintf.h
```diff
@@ -125,22 +567,130 @@ typedef sai_status_t (*sai_set_host_interface_attribute_fn)(
...
+*
+* Return Values:
+*    SAI_STATUS_SUCCESS on success
+*    SAI_STATUS_BUFFER_OVERFLOW if buffer_size is insufficient,
+*    and buffer_size will be filled with required size. Or
+*    if attr_count is insufficient, and attr_count
+*    will be filled with required count.
+*    Failure status code on error
+*/
+typedef sai_status_t(*sai_recv_hostif_packet_fn)(
```
atitjain: Is this for the channel type CB. If that is the case shouldn't adapter host provide this.

sai/inc/saihostintf.h
```diff
@@ -46,23 +483,29 @@ typedef enum _sai_host_interface_attr_t
...
 
-    /* Assosiated port or router interface [sai_object_id_t] 
-     * (MACDATORY_ON_CREATE|CREATE_ONLY) */
+    /* Type [sai_host_interface_type_t] (MANDATORY_ON_CREATE|CREATE_ONLY) */
+    SAI_HOST_INTERFACE_ATTR_TYPE,
+
+    /* Assosiated port or router interface [sai_object_id_t]
+    * Valid only when SAI_HOST_INTERFACE_ATTR_TYPE == SAI_HOST_INTERFACE_TYPE_NETDEV
+    *   (MANDATORY_ON_CREATE when SAI_HOST_INTERFACE_ATTR_TYPE == SAI_HOST_INTERFACE_TYPE_NETDEV | CREATE_ONLY) */
     SAI_HOST_INTERFACE_ATTR_PORT_RIF_ID,
```
atitjain: why to assume it being a router interface. One can use if for L2 protocols as well. The comment mentions it that way but the name is confusing.

lguohan: is SAI_HOST_INTERFACE_ATTR_PORT_OR_RIF_ID ok?

sai/inc/saihostintf.h
lguohan: agree. sai_hostif_trap_group_attr_t seems to be better name

sai/inc/saihostintf.h
```diff
@@ -37,8 +37,445 @@
...
+    SAI_HOSTIF_TRAP_ID_MLD_V2_REPORT = 0x00002015,
+
+    SAI_HOSTIF_TRAP_ID_ROUTER_CUSTOM_RANGE_BASE = 0x0003000,
+
+    /*
+    * pipeline exceptions
+    */
+    SAI_HOSTIF_TRAP_ID_L3_MTU_ERROR = 0x00004000,
+
+    SAI_HOSTIF_TRAP_ID_TTL_ERROR = 0x00004001,
```
lguohan: as we have this one, should we remove ttl violation in switch attribute?

sai/inc/saihostintf.h
mattyk: since the QoS queue definition  is not part of v0.9.2 we can't use it

sai/inc/saihostintf.h
mattyk: i n gereral we can unified all object  function  
IMO keeping then seperate is more clear to the user

sai/inc/saihostintf.h
mattyk: we can define it as optional
i don't a reason why spanning tree application must "know" the LAG port membership

sai/inc/saihostintf.h
mattyk: he only reason we separate user defined trap creation and trap creation was in order to be more user friendly,

sai/inc/saihostintf.h
mattyk: he reason we have send/receive function in SAI is in order to parse/build the SAI packet metadata

sai/inc/saihostintf.h
mattyk: agree . 
SAI_HOSTIF_ATTR_PORT_ID,
 SAI_HOSTIF_ATTR_RIF_ID, ?

sai/inc/saihostintf.h
mattyk: IMO yes we should remove it from the switch attribute

sai/inc/saihostintf.h
lguohan: suggest to combine them as one attribute and use unified object id to tell the object id type.

sai/inc/saihostintf.h
```diff
@@ -35,34 +35,467 @@
...
+    /*
+    * pipeline exceptions
+    * default  action for all pipeline exceptions  is trap
+    */
+    SAI_HOSTIF_TRAP_ID_L3_MTU_ERROR = 0x00004000,
+    
+    /* generated  for Packets with TTL 0 or 1 */
+    SAI_HOSTIF_TRAP_ID_TTL_ERROR = 0x00004001,
+
+    SAI_HOSTIF_TRAP_ID_L3_RPF = 0x00004002,
```
lguohan: sai does not have urpf support now. this exception is better to be removed for now. Will add it back when sai has urpf support.

sai/inc/saihostintf.h
```diff
@@ -35,34 +35,467 @@
...
+
+    SAI_HOSTIF_TRAP_ID_L3_ASSERT = 0x00004003,
```
lguohan: it's unclear what does this l3 assert means. removed for now. will add it back once its meaning becomes clear.

sai/inc/saihostintf.h
```diff
@@ -35,34 +35,467 @@
...
+typedef sai_status_t(*sai_get_hostif_trap_group_attribute_fn)(
+    _In_ sai_object_id_t hostif_trap_group_id,
+    _In_ uint32_t attr_count,
+    _Inout_ sai_attribute_t *attr_list
+    );
+
+typedef enum _sai_hostif_trap_id_t
+{
+    /* default trapid*/
+    SAI_HOSTIF_TRAP_ID_DEFAULT = 0x00000000,
```
lguohan: use the switch attribute to define the default, it's more clear.

sai/inc/saihostintf.h
```diff
@@ -35,34 +35,467 @@
...
+* Arguments:
+*  [in] hostif_trap_id - host interface trap id
+*  [in] attr_count - number of attributes
+*  [in] attr_list - array of attributes
+*
+* Return Values:
+*    SAI_STATUS_SUCCESS on success
+*    Failure status code on error
+*/
+typedef sai_status_t(*sai_create_hostif_trap_fn)(
```
lguohan: remove create/remove_hostif_trap_fn. all these trap id are pre-created in the ASIC, and they have different default value, it makes more sense to have only set/get function for them.

sai/inc/saihostintf.h
mattyk: its OK ,since we don't support l3 multicast

sai/inc/saihostintf.h
```diff
@@ -125,22 +557,124 @@ typedef sai_status_t (*sai_set_host_interface_attribute_fn)(
...
+*
+* Return Values:
+*    SAI_STATUS_SUCCESS on success
+*    Failure status code on error
 */
-typedef struct _sai_host_interface_api_t
+typedef sai_status_t(*sai_send_hostif_packet_fn)(
+    _In_ sai_object_id_t  hif_id,
+    _Inout_ void *buffer,
+    _Inout_ sai_size_t *buffer_size,
```
karthik-krishnamurthy: This might be a typo but shouldn't buffer and buffer_size be just _In_ parameters for send ?

PR #65, opened 2015-04-23
FDB-Flush-API

arunsubash:
Consolidation of FDB Flush API's. Provide options to flush Static/Dynamic/All entries

### DIFF
--- a/sai/inc/saifdb.h
+++ b/sai/inc/saifdb.h
@@ -102,6 +102,52 @@ typedef enum _sai_fdb_entry_attr_t
 } sai_fdb_entry_attr_t;
 
 /*
+*  FDB Flush entry type.
+*/
+typedef enum _sai_fdb_flush_entry_type_t
+{
+    /* Flush dynamic FDB entries */
+    SAI_FDB_FLUSH_ENTRY_DYNAMIC,
+
+    /* Flush static FDB entries */
+    SAI_FDB_FLUSH_ENTRY_STATIC,
+
+    /* Flush all types of FDB entries */
+    SAI_FDB_FLUSH_ENTRY_ALL,
+
+} sai_fdb_flush_entry_type_t;
+
+/*
+* Attribute for FDB delete. These can be used in the following combinations
+* in the flush API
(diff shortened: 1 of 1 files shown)


sai/inc/saifdb.h
```diff
@@ -102,6 +102,52 @@ typedef enum _sai_fdb_entry_attr_t
...
+* Additionally the type of entry needs to be flushed can be specified
+  using SAI_FDB_FLUSH_ATTR_ENTRY_TYPE which can be static or dynamic or all
+*/
+typedef enum _sai_fdb_flush_attr {
+
+   /*Flush based all entries. No value required*/
+   SAI_FDB_FLUSH_ATTR_ALL,
+
+   /*Flush based on MAC [sai_mac_t]*/
+   SAI_FDB_FLUSH_ATTR_MAC,
```
lguohan: do we need this one? this does not appear in the comment.

sai/inc/saifdb.h
```diff
@@ -102,6 +102,52 @@ typedef enum _sai_fdb_entry_attr_t
...
+
+   /*Flush based on port [sai_object_id_t]*/
+   SAI_FDB_FLUSH_ATTR_PORT_ID,
+
+   /*Flush based on VLAN [sai_vlan_id_t]*/
+   SAI_FDB_FLUSH_ATTR_VLAN_ID,
+
+   /*Flush based on entry type [sai_fdb_flush_entry_type_t]*/
+   SAI_FDB_FLUSH_ATTR_ENTRY_TYPE,
```
lguohan: can we re-use the sai_fdb_entry_type_t here? If you do not provide this attribute, then it means all types of entry. Seems to you do not need SAI_FDB_FLUSH_ENTRY_ALL and you can reuse sai_fdb_entry_type_t.

arunsubash: Yes we can reuse the sai_fdb_entry_type_t. Deleted the MAC and ALL option from sai_fdb_flush_attr and updated the pull request

lguohan:
merged

PR #69, opened 2015-04-24
Sample-Flow-Enable

arunsubash:
Adding 2 more attributes to support Flow based sampling - Need to enable sample rate on a port to be used for Flow based sampling. The comments in the header has the details


sai/inc/saiport.h
```diff
@@ -277,6 +277,30 @@ typedef enum _sai_port_attr_t
...
+     * addition NPU's may have different ways to enable port and flow based
+     * Sampling on a given port. The attributes below are expected to abstract 
+     * that and hence the need for having separate attributes for FLOW Based.
+     */
+
+    /* Enable/Disable Flow Based Samplepacket session [sai_object_id_t]. 
+     * Enable ingress sampling by assigning samplepacket object id 
+     * Disable ingress sampling by assigning SAI_NULL_OBJECT_ID as
+     * attribute value */
+    SAI_PORT_ATTR_INGRESS_FLOW_SAMPLEPACKET_ENABLE,
```
atitjain: "The flow that gets sampled is defined by ACL rules", so in this case the ACL should support action SAMPLE. If ACL supports it then we need not have additional port attribute.Given its flow based, there can be some port independent ACLs also,  say a VLAN type router interface. Is it good idea to tie this with port.

Also, the description is little confusing. From the name *_ENABLE it seems to be a bool, but description says "Enable ingress sampling sample packet object id". Let me if I overlooked something.

arunsubash: We already have an ACL Action to Sample. The new attributes are needed to specify the sampling rate to use for a flow depending on the ingress or egress ports. Without this attribute the packet may qualify for sampling but may not have a sample rate configured to actually sample

The reason i added as "Enable" is because NPU's may control sampling for Flow based thru some register though we may have a SAMPLE action

lguohan: Let's first talk a little bit about the model here. SAI allows you create a sample session and define a sample rate for the session. A sample session is sampler. Packets go into this sampler and gets sampled based on a sample rate. You can feed different streams of packets into this sampler. If the different streams have overlapping packets, the overlapped packets are treated as one packet in the sampler. You can create different sample sessions. basically, you create different samplers. One packet can be feed to different samplers, and the sampling process in different different sampler are independent.

Let's now map this model to different hardware. For an ASIC which has only one hardware sampler, then you can only create one sample session, and map port and acl to that sample session. 

For an ASIC which has two hardware samplers, then you can create two sample sessions, and map port to one session, and map acl to the other session. There might be hardware limitation that these two hardware samplers are used for different purpose. For example, one is used for port sampling, the other one is used for acl sampling. In this case, you then cannot map port and acl to a same session.

Let me know if there is any issue with this model. Looks to me these two attributes are not necessary.

arunsubash: As discussed we will do away with these 2 attributes. I am closing the pull request. The summary is

1 - For port the session is enabled using the port based attribute. For flow its enabled for a group of ports thru the SAMPLE ACL Action
2 - On a NPU where there is only one sampler per port supported in HW then we can enable only port based or flow based on a port and not both. 
a.	In case port based is enabled on a port then flow based cannot be enabled for the port unless the port based is disabled and vice versa. 
b.	A given port can be part of multiple Flows that are sampled but it should be using the same sample session. A error would be returned if that is not the case
3 - On a NPU where there is more than one sampler supported in HW such that port based and flow based are separate, the 1st restriction 2a does not apply but the 2nd one 2b would apply

arunsubash:
As per the discussion and conclusion reached i am closing this request

PR #70, opened 2015-04-24
ACL-Size

arunsubash:
Add a new attribute to specify the size of the table. Its optional. Dynamic adjustment of the size depends on NPU support

### DIFF
--- a/sai/inc/saiacl.h
+++ b/sai/inc/saiacl.h
@@ -119,7 +119,21 @@ typedef enum _sai_acl_table_attr_t
      * (default = 0) - Grow dynamically till MAX ACL TCAM Size
      * By default table can grow upto maximum ACL TCAM space. 
      * Supported only during Table Create for now until NPU 
-     * supports Dynamic adjustment of Table size post Table creation */
+     * supports Dynamic adjustment of Table size post Table creation 
+     *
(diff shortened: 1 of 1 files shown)


sai/inc/saiacl.h
```diff
@@ -115,6 +115,12 @@ typedef enum _sai_acl_table_attr_t
      *  SAI_SWITCH_ATTR_ACL_TABLE_MAXIMUM_PRIORITY] */
     SAI_ACL_TABLE_ATTR_PRIORITY,
 
+    /* Table size [sai_uint32_t]
```
lguohan: Please add CREATE_ONLY flag as we do not plan to support set operation on this attribute.

lguohan: Since it's not MANDATORY_ON_CREATE, it should have a default value. By default, we allow the table to grow, therefore the default value is 0.

arunsubash: I will add a comment CREATE_ONLY. Yes i will make a explicit mention that the default value is 0 which indicates grow up to max size. I will resubmit with this change

arunsubash:
Updated the pull request incorporating the comments

lguohan:
merged

PR #77, opened 2015-05-04
Removed all sai obejcts .h included in sai.h

AshokDaparthi:
sai.h files has only the API query functions. Removed  files are no need to include in sai.h. This can be included by .c files where ever  and what ever it required.


tushar-ty:
I think its not a question of good or bad practice but a choice of design and style.

There are several examples of both styles in the public|private domain code, where either only one file needs to be included or all individual files need to be included, and both styles have their pros and cons.

For example I find it cleaner, simpler, abstracted and future proof to just include one file and not care about further details while implementing a module or test code. Unless compilation time is really getting effected or there are unsolvable dependencies or re-entrance issues I don't see this as a bad practice.

So I would propose a hybrid approach which is to create a header file say 'saicommon.h' which includes all sai header files just like sai.h currently does so that users who want to stick to the current style can continue doing the same, i.e include only one file; then you can also go ahead and change sai.h as proposed in this pull request thereby paving way for individual header file includes.

What do you think ?

atitjain:
Couple of thoughts:
1. Should consider renaming sai.h to indicate its functions, currently it sounds like generic sai file
2. Idea of a common file is good, other option is to leave it to the user to create it in his implementation. In the sai headers the inclusion will happen based on the actual relation of the APIs.

tushar-ty:
Should rename the current sai.h to saiinit.h and create a new common sai.h with all the headers - including saiinit.h. 
This way there will be no change required in the current code base.
The purpose of the common header file is to make the usage easy - wherever required.

AshokDaparthi:
I am fine with this .Gouhan do you have nay concerns.

lguohan:
I am fine with tushar's proposal. I think saiapi.h may be a better name than saiinit.h. We can move sai_log_set to sailog.h, and sai_object_type_query to saitypes.h?

AshokDaparthi:
thanks, i will do it and update.

PR #78, opened 2015-05-07
Update spec.md

ben-gale:
First set of proposed changes (through "SAI Functional APIs") - mostly editorial. 

Note that I am also proposing a change to Figure 1 (sent separately), but I don't have the source for this. This change would show a SAI instance per adaptor host, as is required by the 'connect' call.


lguohan:
converted the CRLF -> LF and checked in.

PR #79, opened 2015-05-12
Adding Enum data type in ACL Action to support Packet Action

ramangupta:


sai/inc/saitypes.h
```diff
@@ -258,6 +258,7 @@ typedef struct _sai_acl_action_data_t
       sai_uint16_t u16;
       sai_uint32_t u32;
       sai_mac_t mac;
+      sai_int32_t s32;
```
atitjain: Could you please elaborate the use case a bit more. Packet action is an enum (sai_packet_action_t) so u32 should have been fine. Let me know if I am missing something.

ramangupta: This is as per the spec in the same file where sai_attribute_value_t is defined. Expression that define the value of the enum are of integer data type and therefore signed data type is used. 

/* Data Type to use enum's as attribute value is sai_int32_t s32 */

lguohan:
merged

PR #81, opened 2015-05-12
Qos proposal headers

AshokDaparthi:


sai/inc/sai.h
```diff
@@ -77,6 +82,12 @@ typedef enum _sai_api_t
     SAI_API_SAMPLEPACKET     = 15,  /* sai_samplepacket_api_t */
     SAI_API_STP              = 16,  /* sai_stp_api_t */
     SAI_API_LAG              = 17,  /* sai_lag_api_t */
+    SAI_API_POLICER          = 18,  /* sai_policer_api_t */
+    SAI_API_WRED             = 19,  /* sai_wred_api_t*/
```
atitjain: Have we added APIs for buffers in this?

sai/inc/saipolicer.h
```diff
@@ -0,0 +1,279 @@
...
+    SAI_POLICER_MODE_STORM_CONTROL = 0x00000002,
+
+    /* -- */
+    /* Custom range base value */
+    SAI_POLICER_MODE_CUSTOM_RANGE_BASE = 0x10000000
+} sai_policer_mode_t;
+
+
+/** @brief Enum defining Policer color source */
+typedef enum _sai_policer_color_source_t
```
atitjain: Please consider renaming to color_awareness or something more indicative of the function. Source should be some other table/look-up  in the pipeline that derives the color for packet.

sai/inc/saipolicer.h
```diff
@@ -0,0 +1,279 @@
...
+     */
+    SAI_POLICER_ATTR_METER_TYPE = 0x00000000,
+
+    /** Policer mode [sai_policer_mode_t],
+     * Default SAI_POLICER_MODE_Tr_TCM */
+    SAI_POLICER_ATTR_MODE = 0x00000001,
+
+    /** Policer Color Source [sai_policer_color_source_t],
+     * Default is SAI_POLICER_COLOR_SOURCE_AWARE  */
+    SAI_POLICER_ATTR_COLOR_SOURCE = 0x00000002,
```
atitjain: SAI_POLICER_ATTR_COLOR_SOURCE, please rename to SAI_POLICER_ATTR_COLOR_AWARE

sai/inc/saipolicer.h
```diff
@@ -0,0 +1,279 @@
...
+
+    /* -- */
+    /* Custom range base value */
+    SAI_POLICER_ATTR_CUSTOM_RANGE_BASE = 0x10000000
+
+} sai_policer_attr_t;
+
+
+
+/** @brief Enum defining policer statistics */
```
atitjain: Consider adding discard/drop counts here. For the case when policer action is discard/drop.

sai/inc/saipolicer.h
```diff
@@ -0,0 +1,279 @@
...
+typedef enum _sai_policer_stat_counter_t
+{
+    /** get/set packet count [uint64_t] */
```
atitjain: Enum starting @ 1. Any particular reason?

sai/inc/saiport.h
```diff
@@ -184,6 +184,12 @@ typedef enum _sai_port_attr_t
     /* Current breakout mode [sai_port_breakout_mode_type_t] */
     SAI_PORT_ATTR_CURRENT_BREAKOUT_MODE,
 
+    /* Number of Egress queues per port*/
+    SAI_PORT_ATTR_QOS_NUMBER_OF_QUEUES,
```
atitjain: This is good for readability but as the sai_object_list_t has count as well, this is redundant. Is this the MAX? could we rename if the count and this is different.

sai/inc/saiport.h
```diff
@@ -232,13 +238,17 @@ typedef enum _sai_port_attr_t
        (default to 1514 bytes*/
     SAI_PORT_ATTR_MTU,
 
-    /* [bool] (default to FALSE) */
+    /* Enable flood (unknown unicast or unknown multicast)
+       storm control policer on port [sai_object_id_t]
+       Policer id = SAI_NULL_OBJECT_ID to disable policer on port */
     SAI_PORT_ATTR_FLOOD_STORM_CONTROL,
```
atitjain: Consider changing the name and comments. Comment says bool and then sai_object_id_t both. If its not bool please consider SAI_PORT_ATTR_FLOOD_STORM_CONTROL[_POLICER]_ID. Please consider same for FLOOD and MCAST below.

sai/inc/saiport.h
```diff
@@ -277,6 +287,56 @@ typedef enum _sai_port_attr_t
...
     SAI_PORT_ATTR_EGRESS_SAMPLEPACKET_ENABLE,
 
+    /* Attach/Detach policer to port [sai_object_id_t],
+     * Policer id = SAI_NULL_OBJECT_ID to disable policer on port */
+    SAI_PORT_ATTR_POLICER_ID,
+
+    /* Port default Traffic class Mapping */
+    SAI_PORT_ATTR_QOS_DEFAULT_TC,
+
+   /* Enable DOT1P -> TC MAP [sai_object_id_t] on port
```
atitjain: Not sure if we have a same map on the switch level for the devices that do not support per port mapping.

sai/inc/saiport.h
```diff
@@ -277,6 +287,56 @@ typedef enum _sai_port_attr_t
...
+    SAI_PORT_ATTR_QOS_DSCP_TO_TC_MAP,
+
+   /* Enable DSCP -> COLOR MAP [sai_object_id_t] on port
+    * MAP id = SAI_NULL_OBJECT_ID to disable map on port.
+    * To enable/disbale trust DSCP, Map ID should be add/remove on port */
+    SAI_PORT_ATTR_QOS_DSCP_TO_COLOR_MAP,
+
+   /* Enable TC -> Queue MAP [sai_object_id_t]  on port
+      Map id = SAI_NULL_OBJECT_ID to disable map on port */
+    SAI_PORT_ATTR_QOS_TC_TO_QUEUE_MAP,
```
atitjain: I remember this is discussed offline but for others opinion, should we cosider (TC, Color) to derive the queue.

sai/inc/saiqosqueue.h
```diff
@@ -0,0 +1,251 @@
...
+
+    /** Attach WRED ID to queue [sai_wred_id_t]
+       Mandatory when SAI_QOS_QUEUE_ATTR_DROP_TYPE = SAI_QOS_DROP_TYPE_WRED */
+    SAI_QOS_QUEUE_ATTR_WRED_PROFILE_ID = 0x00000002,
+
+    /** Attach buffer profile to Queue */
+    SAI_QOS_QUEUE_ATTR_BUFFER_PROFILE_ID = 0x00000003,
+
+    /** Attach scheduler to Queue [sai_object_id_t]*/
+    SAI_QOS_QUEUE_ATTR_SCHEDULER_PROFILE_ID = 0x00000004,
```
atitjain: This is profile id or scheduler id itself?

sai/inc/saiqosqueue.h
```diff
@@ -0,0 +1,251 @@
...
+/**
+ * @brief Get attribute to Queue
+ * @param[in] queue_id queue id to set the attribute
+ * @param[in] attr_count number of attributes
+ * @param[inout] attr_list Array of attributes
+ *
+ * @return  SAI_STATUS_SUCCESS on success
+ *           Failure status code on error
+ */
+typedef sai_status_t (*sai_get_qos_queue_attribute_fn)(
```
atitjain: Who manages queue ids, is user supposed to do it?

sai/inc/saischeduler.h
```diff
@@ -0,0 +1,160 @@
...
+/**
+ * @brief  Remove Scheduler profile
+ *
+ * @param[in] scheduler_id Scheduler id
+ *
+ * @return  SAI_STATUS_SUCCESS on success
+ *          Failure status code on error
+ */
+typedef sai_status_t (*sai_remove_scheduler_fn)(
+    _In_ sai_object_id_t scheduler_id
```
atitjain: is this profile or this is scheduler itself?

sai/inc/saiswitch.h
```diff
@@ -170,6 +170,9 @@ typedef enum _sai_switch_attr_t
     /* Default SAI STP instance ID [sai_object_id_t] */
     SAI_SWITCH_ATTR_DEFAULT_STP_INST_ID,
 
+    /** Maximum traffic classes limit*/
+    SAI_SWITCH_ATTR_QOS_MAX_NUMBER_OF_TRAFFIC_CLASSES,
```
atitjain: May be we should add maps here for the devices/implementations that support per switch.

sai/inc/sai.h
AshokDaparthi: No, Mellonax will add it

sai/inc/saipolicer.h
AshokDaparthi: Idea is to extend this later to select the source from which table color has to take like SAI_POLICER_COLOR_SOURCE_DSCP/DOT1p etc. For base version i made it as blind and aware.

sai/inc/saiport.h
AshokDaparthi: Thanks for pointing, i will remove it.

sai/inc/saiport.h
AshokDaparthi: i will change.

sai/inc/saiport.h
AshokDaparthi: Yes i will add it, SWITCH level attributes can be treated as 2 ways. Apply map to switch level if npu support only switch level. Other way instead of apply it on each port we can apply it switch level. We need to consider second should allow in case NPU supports only port level. My take is allow switch attributes and apply it on all ports. If we apply maps at port it will take more precedence. Let me know your views.

sai/inc/saiport.h
AshokDaparthi: For base version tc to queue is enough. We can enhance tc, color later. We will wait for others comments also.

sai/inc/saiqosqueue.h
AshokDaparthi: My proposal scheduler is a profile it contains the values. So it is scheduler profile id.

sai/inc/saiqosqueue.h
AshokDaparthi: SAI will export sai_object_id_t for all queues. application will map to their application queues.

sai/inc/saipolicer.h
AshokDaparthi: Corrected

sai/inc/saipolicer.h
AshokDaparthi: Added

sai/inc/saiport.h
```diff
@@ -277,6 +288,56 @@ typedef enum _sai_port_attr_t
...
+    SAI_PORT_ATTR_QOS_DROP_TYPE,
+
+   /* Attach WRED to port [sai_object_id_t]
+     (mandatory when SAI_PORT_ATTR_QOS_DROP_TYPE =  SAI_QOS_DROP_TYPE_WRED) */
+    SAI_PORT_ATTR_QOS_WRED_ID,
+
+    /* Scheduler for port [sai_object_id_t]*/
+    SAI_PORT_ATTR_QOS_SCHEDULER_ID,
+
+
```
lguohan: We need 
/* attach buffer profile to ingress port */
SAI_PORT_ATTR_INGRESS_BUFFER_PROFILE_ID 

/* attach buffer profile to egress port */
SAI_PORT_ATTR_EGRESS_BUFFER_PROFILE_ID

sai/inc/saischedulergroup.h
```diff
@@ -0,0 +1,192 @@
...
+    /** READ-ONLY */
+    
+    /* Maximum Number of childs on group [uint32_t] */
+    SAI_SCHEDULER_GROUP_ATTR_MAX_SUPPORTED_CHILDS = 0x00000000,
+
+    /** Number of queues/groups childs added to
+     * scheduler group [uint32_t] */
+    SAI_SCHEDULER_GROUP_ATTR_CHILD_COUNT = 0x00000001,
+
+    /** Scheduler Group child obejct id List [sai_object_list_t] */
```
atitjain: Is this going to be the array with size SAI_SCHEDULER_GROUP_ATTR_MAX_SUPPORTED_CHILDS? Is there a reason not to build the hierarchy with each node pointing to a parent. 
1. This models the flow of data in the system in a better way. Queue->parent group...->...-> port.
2. Probably HW tables are also designed that way to enable above, so will reduce SW states.  
3. Enables traversal from child to parent which is not possible in this scheme.

sai/inc/saischedulergroup.h
```diff
@@ -0,0 +1,192 @@
...
+ */
+
+/**
+ * @brief Structure to retrive the scheduler group limits.
+ */
+
+typedef struct _sai_scheduler_group_count_t  {
+
+    _In_ uint32_t level; 
+    _Out_ uint32_t count;
```
atitjain: Is this max or current count? The comment says limit, renaming accordingly will help.

AshokDaparthi: This is h/w global group limit each level. this is not current limit.

sai/inc/saischedulergroup.h
AshokDaparthi: SAI_SCHEDULER_GROUP_ATTR_MAX_SUPPORTED_CHILDS - Maximum childs can be attached to group. 
SAI_SCHEDULER_GROUP_ATTR_CHILD_COUNT - Current/Active childs attachced to group.

AshokDaparthi: We can do that way also. I followed the next hop -> group model. We will take others choice also before deciding.

tushar-ty:
Regarding the QoS maps are we not planning to support mapping of tunneled packets case to traffic class for now ? i.e everything is assumed to be for outer headers as of now ? 
Also as I understand vlan pcp mapping is not being considered for now ?

sai/inc/sai.h
tushar-ty: The following should be deleted from sai.h:
SAI_API_QOS              = 11,  /* sai_qos_api_t */

sai/inc/saiqosqueue.h
```diff
@@ -0,0 +1,251 @@
...
+ *  \{
+ */
+
+/**
+ * @brief Enum defining Queue types.
+ */
+typedef enum _sai_qos_queue_type_t
+{
+     /** H/w Queue for all types of traffic */
+    SAI_QOS_QUEUE_TYPE_NONE = 0x00000000,
```
tushar-ty: How about naming this SAI_QOS_QUEUE_TYPE_ALL or SAI_QOS_QUEUE_TYPE_ANY instead of NONE ?

sai/inc/saiqosqueue.h
```diff
@@ -0,0 +1,251 @@
...
+/**
+ *  @brief Qos methods table retrieved with sai_api_query()
+ */
+typedef struct _sai_qos_api_t
+{
+    sai_set_qos_queue_attribute_fn   set_queue_attribute;
+    sai_get_qos_queue_attribute_fn   get_queue_attribute;
+    sai_get_qos_queue_stats_fn       get_queue_stats;
+
+} sai_qos_api_t;
```
tushar-ty: Naming convention - should be sai_qos_queue_api_t.

sai/inc/saiqosqueue.h
```diff
@@ -0,0 +1,251 @@
...
+    _In_ sai_object_id_t queue_id,
+    _In_ const sai_qos_queue_stat_counter_t *counter_ids,
+    _In_ uint32_t number_of_counters,
+    _Out_ uint64_t* counters
+    );
+
+/**
+ *  @brief Qos methods table retrieved with sai_api_query()
+ */
+typedef struct _sai_qos_api_t
```
tushar-ty: Naming convention - should be _sai_qos_queue_api_t.

sai/inc/sai.h
AshokDaparthi: Done, Keeping the below enum with same value to avoid mismatch with different versions.

sai/inc/saiqosqueue.h
AshokDaparthi: changed to ALL.

sai/inc/saiqosqueue.h
AshokDaparthi: Done.

sai/inc/saiport.h
```diff
@@ -277,6 +288,80 @@ typedef enum _sai_port_attr_t
...
+   /* sai_qos_drop_type_t, Default (TAIL DROP)*/
+    SAI_PORT_ATTR_QOS_DROP_TYPE,
+
+   /* Attach WRED to port [sai_object_id_t]
+     (mandatory when SAI_PORT_ATTR_QOS_DROP_TYPE =  SAI_QOS_DROP_TYPE_WRED) */
+    SAI_PORT_ATTR_QOS_WRED_ID,
+
+    /* Scheduler for port [sai_object_id_t],
+     * Default no limits */
+    SAI_PORT_ATTR_QOS_SCHEDULER_ID,
```
tushar-ty: Why is the scheduler object named differently for queue vs port: SAI_PORT_ATTR_QOS_SCHEDULER_ID, SAI_QOS_QUEUE_ATTR_SCHEDULER_PROFILE_ID ?

sai/inc/saiqosqueue.h
```diff
@@ -0,0 +1,251 @@
...
+
+    /* READ-WRITE */
+
+    /**  Drop Type [sai_qos_drop_type_t],
+       Default  (TAIL DROP) */
+    SAI_QOS_QUEUE_ATTR_DROP_TYPE = 0x00000001,
+
+    /** Attach WRED ID to queue [sai_wred_id_t]
+       Mandatory when SAI_QOS_QUEUE_ATTR_DROP_TYPE = SAI_QOS_DROP_TYPE_WRED */
+    SAI_QOS_QUEUE_ATTR_WRED_PROFILE_ID = 0x00000002,
```
tushar-ty: The WRED object is named differently for queue vs port:  SAI_PORT_ATTR_QOS_WRED_ID, SAI_QOS_QUEUE_ATTR_WRED_PROFILE_ID ?

sai/inc/saischeduler.h
```diff
@@ -0,0 +1,160 @@
...
+ * @brief Enum defining scheduling algorithm.
+ */
+typedef enum _sai_scheduling_type_t
+{
+
+    /* Strict Scheduling */
+    SAI_SCHEDULING_STRICT = 0x00000000,
+
+    /* Deficit Weighted Round-Robin Scheduling */
+    SAI_SCHEDULING_DWRR = 0x00000001,
```
tushar-ty: Could you add WRR before DWRR ?

sai/inc/saiport.h
AshokDaparthi: i will change it

sai/inc/saiqosqueue.h
AshokDaparthi: i will change

sai/inc/saischeduler.h
AshokDaparthi: i will add.

sai/inc/saischeduler.h
```diff
@@ -0,0 +1,160 @@
...
+
+} sai_scheduling_type_t;
+
+/**
+ * @brief Enum defining scheduler attributes.
+ */
+typedef enum _sai_scheduler_attr_t
+{
+    /** Scheduling algorithm [sai_scheduling_type_t ], Default DWRR*/
```
gmirek: The comment suggests the default scheduling mode is DWRR. Since WRR is now being added as a scheduling mode, could we make WRR the default?

AshokDaparthi: I will add it today

sai/inc/saischeduler.h
```diff
@@ -0,0 +1,164 @@
...
+
+
+/**
+ * @brief  Scheduler methods table retrieved with sai_api_query()
+ */
+typedef struct _sai_scheduler_api_t
+{
+    sai_create_scheduler_fn        create_scheduler_profile;
+    sai_remove_scheduler_fn        remove_scheduler_profile;
+    sai_set_scheduler_attribute_fn set_scheduler_attribute;
```
gmirek: In example 2.2.2.1 of SAI-Proposal-Scheduler-Ver4.docx, there are calls to "sai_set_scheduler_attribute_fn()" in which various attributes of an already existing Scheduler profile are getting modified. There are no subsequent calls to apply the newly changed Scheduler via "sai_set_qos_queue_attribute_fn()" or other applicable calls. Does this mean the intent is for the newly changed Scheduler to be applied under the covers of "sai_set_scheduler_attribute_fn()" to all users of this Scheduler(i.e. Ports/Queues)? If so, this would require the QoS Scheduler to maintain a list of all objects attached to the Scheduler. Not sure if that is what is intended here. It seems like the SAI application should take care of maintaining the Scheduler's referred to object list and make the appropraite call accordingly.

AshokDaparthi: Yes, You are correct. We are discussed this approach item in one of the joint meeting with MSFT, BCM and Mellonax. SAI needs to take care of reapplying in case of software profiles.

PR #83, opened 2015-05-19
simplify the return error code related to sai object id: use followin…

lguohan:
simplify the return error code related to sai object id: use following error codes

SAI_STATUS_OBJECT_IN_USE, SAI_STATUS_INVALID_OBJECT_TYPE, SAI_STATUS_INVALID_OBJECT_ID

### DIFF
--- a/sai/inc/saistatus.h
+++ b/sai/inc/saistatus.h
@@ -118,21 +118,6 @@
 #define SAI_MANDATORY_ATTRIBUTE_MISSING     SAI_STATUS_CODE(0x0000000EL)
 
 /*  
-*  Invalid interface id 
-*/
-#define SAI_STATUS_INVALID_INTERFACE        SAI_STATUS_CODE(0x0000000FL)
-
-/*  
-*  Port is in use 
-*/
-#define SAI_STATUS_PORT_IN_USE              SAI_STATUS_CODE(0x00000010L)
-
-/*  
-*  Invalid switch ID 
-*/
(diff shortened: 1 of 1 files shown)


sai/inc/saistatus.h
```diff
@@ -178,6 +138,14 @@
...
 */
 #define SAI_STATUS_INVALID_OBJECT_TYPE      SAI_STATUS_CODE(0x0000001AL)
+
+
+/*
+* Invalid SAI Object ID passed to a function. Return SAI_STATUS_INVALID_OBJECT_TYPE
+* when the object type is invalid, otherwise return this SAI_STATUS_INVALID_OBJECT_ID.
+* This can happen if the object id is already deleted.
+*/
+#define SAI_STATUS_INVALID_OBJECT_ID        SAI_STATUS_CODE(0x0000001BL)
```
SathishKumarTR-Dell: We return the SAI_STATUS_ITEM_NOT_FOUND status for this example - object id is already deleted. If not needed, please remove ITEM_NOT_FOUND and lets retain ITEM_ALREADY_EXISTS for the duplicate creation.

lguohan: I think ITEM_NOT_FOUND and ITEM_ALREADY_EXISTS should both be retained. They are used for create_route/delete_route which are not identified by object id.

ramangupta: Can this pull request be closed as it has been already merged to the master?

PR #84, opened 2015-05-20
Minor Change in SAI ACL and STP file

ramangupta:
Summary of Changes Proposed:
1) Generic Forward action in the ACL action enum has been removed as it is a part of Packet Action enum (SAI_PACKET_ACTION_FORWARD).
2) Comment added in sai_create_stp_fn() to indicate that the Default port state while creating STP instance would be Forwarding


lguohan:
merged

PR #91, opened 2015-05-28
add hash enhancement with udf header files

lguohan:


lguohan:
update udf hash spec. 

Swapped the creation order for udf and udf_group. Now, udf_group is created first to obtain a udf_group_id, then use the udf_group_id to create a udf. This allows to put multiple udfs in a same udf group.

PR #95, opened 2015-06-01
Add neighbor attribute SAI_NEIGHBOR_ATTR_NO_HOST_ENTRY

SathishKumarTR-Dell:
Added a new attribute 'SAI_NEIGHBOR_ATTR_NO_HOST_ENTRY' to SAI neighbor entry. This can be used to indicate to SAI adapter to not program the L3 forwarding entry for the neighbor in ASIC and just cache the neighbor information.


tushar-ty:
This would be the first instance of the specification explicitly indicating the implementation to be stateful. Could the implementation not derive this conclusion on its own ?

SathishKumarTR-Dell:
Dell - Internal Use - Confidential
Thanks for sharing your views.

For IPv6 link-local address neighbor entry, L3 host entry for the IPv6 address should not be programmed in the ASIC – as L3 forwarding for this destination address is not expected.
However, it is valid to create a next-hop for the IPv6 link-local address and point LPM routes to it – for which the implementation needs to associate the neighbor dmac and dport. So, the implementation needs to cache the neighbor entry anyway.

The proposed attribute can be used for any other use-cases like this.

Also, I feel deriving based on the attribute from user instead of the implicit logic could result in uniform interface across different implementations.

Regards,
Sathish

atitjain:
The link local Host use case that you mentioned here, the entry is not programmed in ASIC, now if this is not required to be there in the ASIC as in DATAPATH (let me know if this is not a correct mapping) then why the user need to call SAI API at all which is used to abstract ASIC/DP. 

For the LPM route case again LPM is also used to forward packets. I am missing something here. Please clarify.

SathishKumarTR-Dell:
L3 forwarding entry (host entry) for the link-local address alone should not be installed in ASIC L3 host table. But, the link-local address can be used as a nexthop for routes by the routing protocols. For installing those routes, dmac, dport/vlan of the nexthop needs to be determined from the neighbor entry in SAI implementation.

lguohan:
I feel SAI_NEIGHBOR_ATTR_NO_HOST_ROUTE is probably more clear than HOST_ENTRY, agree?

SathishKumarTR-Dell:
Yes, renamed the attribute now.

lguohan:
merged

Issue #94, opened 2015-06-01
SAI_STATUS_INVALID_VLAN_ID should be more informative


soramichi:
SAI_STATUS_INVALID_VLAN_ID is the only one error code for vlan_id related errors,
but it must be separated into two to provide more information to users.

    #define SAI_STATUS_VLAN_ID_IN_USE
    #define SAI_STATUS_VLAN_ID_OUT_OF_RANGE

The first one is to notify that the given vlan_id is already in use by another VLAN,
and the second one is to notify that the given vlan_id is larger than 4095.

SAI_STATUS_VLAN_ID_OUT_OF_RANGE is useful because sai_vlan_id_t has 16 bits,
thus users cannot see from the type that it must not exceed 4095.
(So it might be another solution to use something like uint12_t, if it can be implemented with low cost.)

tushar-ty:
Shouldn't the user be aware of the standards and not use an invalid value in the first place ?
And shouldn't SAI_STATUS_INVALID_PARAMETER be enough ?

I think that for param values which are specified by standards, there is no need for specific error codes. We need specific error codes for things that are not standard and are vendor specific like SAI_STATUS_INVALID_VRID or SAI_STATUS_INVALID_ACL_TABLE_ID. For that matter even SAI_STATUS_INVALID_VLAN_ID is an overkill although SAI_STATUS_VLAN_ID_IN_USE makes sense as it can't be covered by SAI_STATUS_OBJECT_IN_USE.

soramichi:
Thank you for the kind and quick response.

I understand your point and made the pull request #96 according to the discussion here.

PR #96, opened 2015-06-02
replace SAI_STATUS_INVALID_VLAN_ID by SAI_STATUS_VLAN_ID_IN_USE

soramichi:
Refactored error codes related to vlan in response to the discussion in #94 .

Sample usage of the error codes follows:

    create_vlan(10);    /* OK, returns SAI_STATUS_SUCCESS */
    create_vlan(10000); /* NG, returns SAI_STATUS_INVALID_PARAMETER. The user should know the standards and see 10000 is too large. */
    create_vlan(10);    /* NG, returns SAI_STATUS_VLAN_ID_IN_USE, which is newly introduced in this commit */
    remove_vlan(20);    /* NG, returns SAI_STATUS_INVALID_PARAMETER */


lguohan:
prefer to use SAI_STATUS_ITEM_ALREADY_EXISTS and SAI_STATUS_ITEM_NOT_FOUND for the 3rd and 4th case. 

I prefer to use general error code instead of specific one as long as it can help user to identify the problem. Not sure if SAI_STATUS_INVALID_VLAN_ID is not needed at all.

PR #97, opened 2015-06-02
New SAI ACL Action Proposed to set CPU Queue

ramangupta:
The CPU is responsible for processing control plane traffic. Control plane traffic consists of transmit/receive protocol packets or CPU bound traffic. 

Packets are checked against CAM entries and a packet may be identified as CPU bound based on protocols configured within the system. These CPU bound packets needs to be marked for certain ingress CPU Queues for processing. Each of these CPU Queues is rate limited and would help in avoiding protocol flaps due to heavy CPU bound traffic. 

An additional ACL entry action is needed to specify the CPU queue to which packets need to be sent to the CPU.


sai/inc/saiacl.h
```diff
@@ -450,8 +450,11 @@ typedef enum _sai_acl_entry_attr_t
     /* Set egress packet sampling (samplepacket session id) [sai_object_id_t] */
     SAI_ACL_ENTRY_ATTR_ACTION_EGRESS_SAMPLEPACKET_ENABLE,
 
+    /* Set CPU Queue for CPU bound traffic [sai_uint16_t] */
+    SAI_ACL_ENTRY_ATTR_ACTION_SET_CPU_QUEUE,
```
atitjain: - This action is dependent on the sai_packet_action_t. Just wondering is there a way to club this with sai_packet_action_t. This will be helpful in other cases as well where the pkt cmd can be enabled. 
- Other thing is we have trap id and related parameters defined in saihostintf.h, is there a way to correlated these together. Or should we think of having trap-id as an action rather the CPU queue so that all the trap-id attributes are inherited for the trap through ACL.
- Please check following as well
344     /* ACL traps */
345     SAI_HOSTIF_USER_DEFINED_TRAP_ID_ACL_MIN = 0x00000002,
346 
347     SAI_HOSTIF_USER_DEFINED_TRAP_ID_ACL_MAX = SAI_HOSTIF_USER_DEFINED_TRAP_ID_ACL_MIN + SAI_HOSTIF_USER_DEFINED_TRAP_ID_TABLE_RANGE,

lguohan: In saihostintf.h, there is SAI_HOSTIF_TRAP_GROUP_ATTR_QUEUE to set the cpu egress queue. How is this ACL action related to it?

ramangupta: - sai_packet_action_t has drop and forward too, where this action wouldn't be appropriate. Also packet action is itself a enum attribute value and CPU Queue value cannot be derived from it. 
- If trap-id is added as an action, it should also be qualified. The Table should include all qualifiers which could be used to trap any control packet. 
- All NPUs support control packet trapping mechanism and SAI_HOSTIF_TRAP_GROUP_ATTR_QUEUE is used just to abstract NPUs capability to enqueue control packets to CPU based on trap-id.

atitjain: What about followings, 

344 /* ACL traps */
345 SAI_HOSTIF_USER_DEFINED_TRAP_ID_ACL_MIN = 0x00000002,
346 
347 SAI_HOSTIF_USER_DEFINED_TRAP_ID_ACL_MAX = SAI_HOSTIF_USER_DEFINED_TRAP_ID_ACL_MIN + SAI_HOSTIF_USER_DEFINED_TRAP_ID_TABLE_RANGE,

Are they relevant looking at the explanation above?

ramangupta: I think they have just been defined to identify whether an ACL entry has been hit or not.

arunsubash: Hi Atit,

ACL is sai_hostif_user_defined_trap_id_t is a user defined trap. There are only 2 attributes defined for the user defined trap – TRAP_CHANNEL and FD. For user defined traps the Q is expected to be set as part of the user defined trap in the corresponding table. Hence there is a separation between the user defined traps and general traps

-Arun

lguohan: IIRC, if we want to trap a packet using ACL, we set SAI_ACL_ENTRY_ATTR_PACKET_ACTION to either SAI_PACKET_ACTION_TRAP or SAI_PACKET_ACTION_LOG. Then, we can use the proposed SET_CPU_QUEUE to set the CPU queue. 

Should we also add an action SAI_ACL_ENTRY_ATTR_ACTION_TRAP_ID to set the corresponding trap id for this acl? I did not find it in current list.

ramangupta: Yes, this action needs to be added. Currently the user defined trap id ACL range is just one as both min and max are equal (SAI_HOSTIF_USER_DEFINED_TRAP_ID_TABLE_RANGE = 0x0) 
Trap Id action can be added along with the changes needed to define this range.

lguohan: it seems like this RANGE is set to zero deliberately so that all ACL entries are mapped a single trap_id SAI_HOSTIF_USER_DEFINED_TRAP_ID_ACL_MIN, so current there is no need to define this acl action to set trap for now.

lguohan: in the qos proposal, we have TC_2_EGRESS_QUEUE mapping, is it ok to use TC instead of set the CPU queue directly? the current SAI_ACL_ENTRY_ATTR_ACTION_SET_TC is for forwarded packet, is it ok to use SAI_ACL_ENTRY_ATTR_ACTION_SET_CPU_TC instead of SAI_ACL_ENTRY_ATTR_ACTION_SET_CPU_QUEUE here?

atitjain: What I understand is TC is the Traffic class and typically user will have some 8/16 classes, which would be mapped to the internal queues (8/16). CPU queues can be many more 48/64/128 etc so TC I think cant be used.

lguohan: you are right TC is the traffic class and typically it has 8/16 classes. The way I am looking at is that  try to make the SAI model simple first. Packet comes in, mapped to a TC based on qos_map or whatever mechanism, and then mapped to a egress queue based on the TC2Queue mapping. If we can first make this simple model work on different hardware platforms, that's would be a win for SAI. As you indicated, this may not expose the full hardware capability on all platforms. If customer feels that's not enough, we can address them in the later stage. Usually, we do not use all the 48 cpu queues. At least for our applications, we do not classify our control plane traffic into that many classes.

The question here is if SET_CPU_TC works or not. Here I assume there will be TC_TO_QUEUE mapping for the cpu port.

atitjain: I am ok with above, few points to add 
1. We may have TC to queue mapping per switch as well in some devices, might be rare case  
2. The CPU queues are also used as a mean for segregation of different protocols to different queues and shape them in isolation based on the system scale so that they dont hog each other.

lguohan: Ok. as the reason stated above, I am ok with SET_CPU_QUEUE here. However, can we get more clarification on this SET_CPU_QUEUE for different types of packet actions. 
1. when packet action == SAI_PACKET_ACTION_TRAP, the SET_CPU_QUEUE set the cpu queue for the packet.
2. when packet action == SAI_PACKET_ACTION_LOG, the SET_CPU_QUEUE set the cpu queue for the packet goes to CPU, and SET_TC_QUEUE set the TC for the copy of packet goes to the front panel.
3. when packet action == SAI_PACKET_ACTION_FORWARD and the destination is CPU, e.g., bgp control plane traffic. Does this SET_CPU_QUEUE still work or we should use SET_TC?

ramangupta: We can mandate the usage of SET_CPU_QUEUE to be used only when packet action is either Trap or log.

atitjain: I am still not clear about the use case of SAI_HOSTIF_USER_DEFINED_TRAP_ID_ACL_MIN etc.

lguohan: @atitjain. say you want to use ACL to trap a packet to CPU, then you set SAI_ACL_ENTRY_ATTR_PACKET_ACTION = TRAP for that ACL entry. Since SAI defines several host interfaces (receive func callback, netdev, file descriptor), we need to also specify which host interface the packet goes to. This is done by set up the SAI_HOSTIF_USER_DEFINED_TRAP_ATTR_TRAP_CHANNEL for the SAI_HOSTIF_USER_DEFINED_TRAP_ID_ACL_MIN. 

Current only one ACL trap id is defined, so that all ACL traps share this trap id. In the future, we should have more ACL trap id, each may corresponds to different trap channel. At that time, we need add ACL action to specify which trap ID this ACL points to.

arunsubash: Yes @atitjain that is the plan. When we have more user defined ACL trap ID's then have the trap ID as an ACL action

atitjain: In case of user defined trap attributes, why the attributes are different than the default defined traps. If we make the trap attributes common then just defining the trap Id will solve the CPU queue/priority allocation as well. Also it seems right now for a packet trapped from the ACL we cant mention the channel type if trap id is not used. 

Also in the sai_hostif_trap_attr_t should we have the packet action? We have packet action used in many places (route, neighbor and FDB), these are in the catogory of user defined I think. Now how do I select channel etc for these or for these we expect the CHANNEL_CB alone as channel.

sai/inc/saiacl.h
```diff
@@ -136,6 +136,29 @@ typedef enum _sai_acl_table_attr_t
      * allocated and able to do ACL CAM Carving accurately */
     SAI_ACL_TABLE_ATTR_SIZE,
 
+    /* Table Group Id [sai_uint32_t], CREATE_ONLY
+     * (default = 0) - By default, no grouping will be performed and all tables
+     * will be looked up independently.
+     *
+     * ACL Tables can be grouped together. If any ACL entry is hit within the 
+     * group, the lookup would jump to the next subsequent ACL table which is 
+     * not part of the group.
```
lguohan: "the lookup would jump to the next subsequent ACL table which is not part of the group". This implies the ACL table is looked up sequentially which may not be the case. Since we do not differentiate sequential or parallel lookup here. How about simply focusing on the result of the lookup? For example, 

Within one ACL group, only one ACL entry will be hit which is based on the table priority as well as ACL entry priority within the table. Every ACL Table is associated to a priority which is a mandatory attribute needed at the time of table creation. The order of lookup within the ACL table group will be based on individual table priority. Action(s) from the ACL entry residing in the higher priority table within the group will be performed.

Across the ACL groups, one hit ACL entry from each group will be selected and all non-conflicting actions from them will be performed.

sai/inc/saiacl.h
```diff
@@ -136,6 +136,29 @@ typedef enum _sai_acl_table_attr_t
...
+     *
+     * The ACL Table lookup could be done serially or in parallel. In both the
+     * cases there could be a need to group multiple tables so that only single
+     * ACL rule entry actions are performed.
```
lguohan: This para looks to me more like the motivation to introduction the table group concept. Maybe we can put it to the beginning.

sai/inc/saiacl.h
```diff
@@ -450,8 +473,14 @@ typedef enum _sai_acl_entry_attr_t
     /* Set egress packet sampling (samplepacket session id) [sai_object_id_t] */
     SAI_ACL_ENTRY_ATTR_ACTION_EGRESS_SAMPLEPACKET_ENABLE,
 
+    /* Set CPU Queue for CPU bound traffic [sai_object_id_t]
+     * Action can be used whenever packet is destined to CPU such as
+     * when packet action specifies the packet needs to be punted
+     * to CPU (Trap/Log) or the destination port points to CPU. */
+    SAI_ACL_ENTRY_ATTR_ACTION_SET_CPU_QUEUE,
```
lguohan: this has been merged. can we create a new pull request for the acl group id and close this one? Right now, the thread title is about set cpu queue.

ramangupta: Sure, I will close this request and raise a new one

sai/inc/saiacl.h
```diff
@@ -136,6 +136,29 @@ typedef enum _sai_acl_table_attr_t
...
+     * Without grouping, all non-conflicting actions from different ACL entry
+     * hits across the tables will be performed.
+     *
+     * Every ACL Table is associated to a priority which is a mandatory attribute
+     * needed at the time of table creation. The order of lookup within the ACL
+     * table group will be based on individual table priority. Action(s) from the
+     * ACL entry residing in the higher priority table within the group will be
+     * performed.
+     */
+    SAI_ACL_TABLE_ATTR_GRP_ID,
```
lguohan: another question the group priority. In case of multiple hits from different drop and the action conflicting. For example, acl 1 from group A set cpu queue to 10, acl 2 from group B set cpu queue to 20. which one wins?

sai/inc/saiacl.h
```diff
@@ -136,6 +136,29 @@ typedef enum _sai_acl_table_attr_t
...
+     * ACL Tables can be grouped together. If any ACL entry is hit within the 
+     * group, the lookup would jump to the next subsequent ACL table which is 
+     * not part of the group.
+     *
+     * The ACL Table lookup could be done serially or in parallel. In both the
+     * cases there could be a need to group multiple tables so that only single
+     * ACL rule entry actions are performed.
+     *
+     * Without grouping, all non-conflicting actions from different ACL entry
+     * hits across the tables will be performed.
```
lguohan: another question the group priority. In case of multiple hits from different drop and the action conflicting. For example, acl 1 from group A set cpu queue to 10, acl 2 from group B set cpu queue to 20. which one wins? what's the criteria?

ramangupta: Created a new pull request for ACL Table Group Id #108 from where we can take the discussion forward.

PR #100, opened 2015-06-11
Adding sai_samplepacket_type_t to sai_attribute_value_t 

atitjain:
Adding sai_samplepacket_type_t to sai_attribute_value_t  so that attribute could be set for samplepacket API.


sai/inc/saitypes.h
```diff
@@ -337,6 +337,7 @@ typedef union {
     sai_acl_field_data_t aclfield;
     sai_acl_action_data_t aclaction;
     sai_port_breakout_t portbreakout;
+    sai_samplepacket_type_t sflowtype;
```
lguohan: sai_samplepacket_type_t is a enum, then we should just use sai_int32_t s32 to pass the value.

atitjain: I see, let me check. Thanks.

lguohan:
can we close this issue now?

atitjain:
Closing.

PR #101, opened 2015-06-16
minor changes in saifdb.h, saistp.h and saihostintf.h

atitjain:


sai/inc/saifdb.h
```diff
@@ -94,6 +94,9 @@ typedef enum _sai_fdb_entry_attr_t
     /* FDB entry packet action [sai_packet_action_t] (MANDATORY_ON_CREATE|CREATE_AND_SET) */
     SAI_FDB_ENTRY_ATTR_PACKET_ACTION,
 
+    /* Packet trap-id used for trap/log actions [uint8_t]*/
+    SAI_NEIGHBOR_ATTR_TRAP_ID,
```
lguohan: SAI_FDB_ATTR_TRAP_ID? 

Can we add a default value for this attribute so that it is not MANDATORY_ON_CREATE. The default value can be SAI_HOSTIF_USER_DEFINED_TRAP_ID_FDB_MIN, and the value type can be sai_hostif_user_defined_trap_id_t.

I think we should also allow CREATE_AND_SET for this attribute.

sai/inc/saistp.h
```diff
@@ -60,7 +60,7 @@ typedef enum _sai_stp_attr_t
     /** READ-ONLY */
 
     /** Vlans attached to STP instance [sai_vlan_list_t] */
-    SAI_STP_ATTR_VLAN,
+    SAI_STP_ATTR_VLAN_LIST,
```
lguohan: I have fixed this and pushed to the master. thanks.

sai/inc/sairoute.h
```diff
@@ -43,9 +43,8 @@ typedef enum _sai_route_attr_t
        (default to SAI_PACKET_ACTION_FORWARD) */
     SAI_ROUTE_ATTR_PACKET_ACTION,
 
-    /* Packet priority for trap/log actions [uint8_t] 
-       (default to 0) */
-    SAI_ROUTE_ATTR_TRAP_PRIORITY,
```
lguohan: since the user defined attribute does not priority, I am not sure we should delete the trap priority attribute id here. 
However, should we have trap priority per trap Id or per route? @mattyk

sai/inc/saifdb.h
atitjain: Agree, that was a miss.

sai/inc/sairoute.h
atitjain: I have commented it somewhere I remember. I think we should have trap prio and queue etc for user traps also. That makes the implementation clean. For all the sources from where a packet is trapped/mirrored to CPU all we will need is a trap id.

PR #103, opened 2015-06-19
Modifying comments for system flush

dgsudharsan:


lguohan:
Thanks. Removed SAI_FDB_FLUSH_ENTRY_ALL as it is redundant. Also modified the comment to make it more clear.

atitjain:
Hi Gouhan,

This is not redundant. We need this. This is sai_fdb_flush_entry_type_t which is Dynamic, static or all (SAI_FDB_FLUSH_ENTRY_ALL). Probably rename to SAI_FDB_FLUSH_ENTRY_BOTH. If we remove this, we don’t have a way to flush both static and dynamic in one call.

Regards,
Atit

atitjain:
Like the other two attributes, do not use this entry_type attribute when you want to flush both static and dynamic entries.

From: Jain, Atit
Sent: Saturday, June 20, 12:57 AM
Subject: RE: [OCP-Networking-Project-Community-Contributions] Modifying comments for system flush (#103)
To: opencomputeproject/OCP-Networking-Project-Community-Contributions, opencomputeproject/OCP-Networking-Project-Community-Contributions
Cc: Guohan Lu

Hi Gouhan,



This is not redundant. We need this. This is sai_fdb_flush_entry_type_t which is Dynamic, static or all (SAI_FDB_FLUSH_ENTRY_ALL). Probably rename to SAI_FDB_FLUSH_ENTRY_BOTH. If we remove this, we don't have a way to flush both static and dynamic in one call.



Regards,
Atit

atitjain:
Can we add this as a comment.

dgsudharsan:
I suppose if we plan to do so, we can use sai_fdb_entry_type_t and remove  sai_fdb_flush_entry_type_t. There will also be a case where for system wide static and dynamic flush we might not set any attributes at all. Is it ok to call the flush API with NULL and attr_count as 0?

lguohan:
@atitjain , add comments to further explain how to use SAI_FDB_FLUSH_ATTR_ENTRY_TYPE. @dgsudharsan , not sure if sai_fdb_entry_type_t and sai_fdb_flush_entry_type_t will always be the same. For this purpose, it was designed to be separated.

PR #104, opened 2015-06-22
Add storage layer codes to stub_sai_vlan.c

soramichi:
As a first trial step of contributing storage layer codes to `sai/stub` discussed [here](http://lists.opencompute.org/pipermail/opencompute-networking/2015-June/000926.html),
this pull request enables useful error messaging by comparing given API parameters and the VLAN states already created by former API calls.

For the details of the messaging please refer to each commit. 

What we did is separating VLAN-related codes of our project into three layers (stub, storage, data plane) and
merging the storage layer into `sai/stub` with a bit of modification to adapt to the existing codes.

We would like to get any comments from the community and fix the way we contribute (if necessary) in the early stage of our work, thus we decided to make a pull request although the patch is still small.


itaibaz:
Hi,
Thanks for the contribution

It looks good. Few comments :

1)      Re vlan key - char* id. Perhaps it would be better to use the key as defined in SAI sai_vlan_id_t vlan_id and not the display string

2)      Handling of default vlan ID -

    /* Default VlanID for ports that are not members of

       any vlans [sai_vlan_id_t]  (default to vlan 1)*/

   SAI_SWITCH_ATTR_DEFAULT_PORT_VLAN_ID,
                Initially all the ports should be members of the default vlan ID, so you might want to add it to the DB on switch init

3)      Add Vlan ID range validation – 1 to 4094

Itai

lguohan:
the vlan id range should be 1 to 4095?

soramichi:
Thank you very much for the comments.
I fix the easy ones (vlan key type, range validation) and pushed the changes.

As for the default van id, we are not sure how to neatly implement it to the stub because:
 1. "All ports" cannot be defined without running SAI on an actual switch.
 2. Also, `SAI_SWITCH_ATTR_PORT_LIST` is read-only, thus the user should not set it to emulate the number of ports the user wants for testing.

Changing the abstraction view point from vlan to port and creating a structure like

    struct __port {
        sai_vlan_port_t port;
        sai_vlan_id_t vlan_to_which_this_port_belongs;
    }

allows to set the default value of the second member to `SAI_SWITCH_ATTR_DEFAULT_PORT_VLAN_ID`, but it doesn't solve the problem that "all ports" is not defined.

What do you think about this matter?

soramichi:
Hi lguohan, isn't 4095 (0xFFF) reserved and users must not be able to use it?

atitjain:
I remember in almost all the implementation I have worked on, 4095 is used as an internal VLAN. While its true that it invalid on network, its used inside the system for say communicating to FPGAs, CPU etc. So user should be able to create it and use it internally. If SAI doesnt support it, the internal use is also not possible.

itaibaz:
Hi,
Re port list, I think at this stage we can write a simple implementation, and allow the user to customize it
I would add in SAI switch implementation :

#define PORT_NUMBER 10
Int port_ids[PORT_NUMBER] = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10}

In SAI_SWITCH_ATTR_PORT_LIST getter
Foreach port, add to list  stub_create_object(SAI_OBJECT_TYPE_PORT, port_id[i])

And then you can use it in Vlan

Or alternatively, just skip it in the first stage

Itai

soramichi:
Hi,

thanks again for the comments.

Is it possible to just skip the default VLAN matter now and have this PR merged?
(I'll fix the VLAN id range if necessary.)

We are working on separating other API codes as well, and it looks reasonable to deal with the default VLAN matter in combination with separating/contributing switch API codes.

Soramichi

itaibaz:
Hi
Yes, we can definitely skip the default VLAN for now

Itai

lguohan:
Thanks. looks good. applied to master. squashed into one commit and fix the identation and padding.

PR #106, opened 2015-06-24
Modified port oper status and port event notification to include a list of notifications

kapiladhikesavalu:
Modified port oper status and port event notification to include a list of notifications

Changes for notifying hardware lane list as part of port event notification, which avoids
the additional SAI_PORT_ATTR_HW_LANE_LIST port attribute call after each port event notification.

### DIFF
--- a/sai/inc/saiport.h
+++ b/sai/inc/saiport.h
@@ -151,8 +151,7 @@ typedef enum _sai_port_event_t
 } sai_port_event_t;
 
 /*
- * Defines the port event notification and its
- * associated hardware lane list
+ * Defines the port event notification
  */
 typedef struct _sai_port_event_notification_t {
 
@@ -161,9 +160,6 @@ typedef struct _sai_port_event_notification_t {
 
     /* Port event */
     sai_port_event_t port_event;
-
-   /* Port hardware lane list - [SAI_PORT_ATTR_HW_LANE_LIST] */
-    sai_u32_list_t u32list;
 
 } sai_port_event_notification_t;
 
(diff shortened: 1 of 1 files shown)


sai/inc/saiport.h
```diff
@@ -414,15 +444,15 @@ typedef sai_status_t (*sai_get_port_stats_fn)(
...
+*   [in] port_status - pointer to port operational status
 *
 * Return Values:
 *    None
 */
 typedef void (*sai_port_state_change_notification_fn)(
-    _In_ sai_object_id_t port_id,
-    _In_ sai_port_oper_status_t port_state
+    _In_ uint32_t count,
+    _In_ sai_port_oper_status_notification_t *port_status
```
lguohan: what's the motivation to change to the batch API? Is this related to a performance issue here?

arunsubash: Yes its gives an option for SAI to send multiple events in one shot. It could give performance improvement depending on the system and during init time where all ports could come up at the same time

sai/inc/saiport.h
```diff
@@ -138,6 +151,23 @@ typedef enum _sai_port_event_t
...
+typedef struct _sai_port_event_notification_t {
+
+    /* Port id */
+    sai_object_id_t port_id;
+
+    /* Port event */
+    sai_port_event_t port_event;
+
+   /* Port hardware lane list - [SAI_PORT_ATTR_HW_LANE_LIST] */
+    sai_u32_list_t u32list;
```
lguohan: explain the use case for this attribute a little. For the port create event, app can do a query for the hw lane list after getting this notification. For the port delete event, assume app already maintains a hw lane list mapping.

arunsubash: This is to avoid an additional call to SAI to get the HW Lane mapping on getting a PORT_ADD event. Its more a optimization to reduce the number of calls

lguohan:
merged, thanks.

PR #107, opened 2015-06-24
add SAI default virtual router id

lguohan:
SAI creates default virtual router id at sai_initiate_switch


SathishKumarTR-Dell:
This can be useful for associating with a default id in other objects/attributes. Is there any other specific motivation? I think removing the default vr object id should be not allowed and probably SAI_STATUS_OBJECT_IN_USE can be returned. Is this fine?

lguohan:
yes, it is mainly used to associate with other objects such as route, router interface with this default vr id, so that user does not have to create a vr. Also, for those object such as router interface, l3 tunnel ingress object, we can by default associate this default vr with them.

you are right, it cannot be deleted and should retrn IN_USE. Will add this to the comment.

PR #108, opened 2015-06-24
SAI ACL Table Group Id attribute added

ramangupta:


sai/inc/saiacl.h
```diff
@@ -136,6 +136,29 @@ typedef enum _sai_acl_table_attr_t
      * allocated and able to do ACL CAM Carving accurately */
     SAI_ACL_TABLE_ATTR_SIZE,
 
+    /* Table Group Id [sai_uint32_t], CREATE_ONLY
+     * (default = 0) - By default, no grouping will be performed and all tables
+     * will be looked up independently.
+     *
+     * ACL Tables can be grouped together. If any ACL entry is hit within the
+     * group, the lookup would jump to the next subsequent ACL table which is
+     * not part of the group.
```
lguohan: "the lookup would jump to the next subsequent ACL table which is not part of the group". This implies the ACL table is looked up sequentially which may not be the case. Since we do not differentiate sequential or parallel lookup here. How about simply focusing on the result of the lookup? For example, 

Within one ACL group, only one ACL entry will be hit which is based on the table priority as well as ACL entry priority within the table. Every ACL Table is associated to a priority which is a mandatory attribute needed at the time of table creation. The order of lookup within the ACL table group will be based on individual table priority. Action(s) from the ACL entry residing in the higher priority table within the group will be performed.

Across the ACL groups, one hit ACL entry from each group will be selected and all non-conflicting actions from them will be performed.

sai/inc/saiacl.h
```diff
@@ -136,6 +136,29 @@ typedef enum _sai_acl_table_attr_t
...
+     *
+     * The ACL Table lookup could be done serially or in parallel. In both the
+     * cases there could be a need to group multiple tables so that only single
+     * ACL rule entry actions are performed.
```
lguohan: This paragraph looks to me more like the motivation to introduction the table group concept. Maybe we can put it to the beginning.

sai/inc/saiacl.h
```diff
@@ -136,6 +136,29 @@ typedef enum _sai_acl_table_attr_t
...
+     *
+     * Without grouping, all non-conflicting actions from different ACL entry
+     * hits across the tables will be performed.
```
lguohan: another question the group priority. In case of multiple hits from different drop and the action conflicting. For example, acl 1 from group A set cpu queue to 10, acl 2 from group B set cpu queue to 20. which one wins? what's the criteria?

ramangupta: The behavior would be similar to the non-grouping case. Action(s) from the highest table priority will be taken. Grouping is just to avoid picking actions from multiple hits between tables with different qualifier set. In the case mentioned above, irrespective of the groups, cpu queue will be taken from the entry that belongs to higher ACL table priority

sai/inc/saiacl.h
ramangupta: I'll re-frame the comments and push the changes.

sai/inc/saiacl.h
```diff
@@ -136,6 +136,33 @@ typedef enum _sai_acl_table_attr_t
...
+     *
+     * Within one ACL group, only one ACL entry will be hit which is based on the 
+     * table priority as well as ACL entry priority within the table. Every ACL 
+     * Table is associated to a priority which is a mandatory attribute needed at 
+     * the time of table creation. The order of lookup within the ACL table group 
+     * will be based on individual table priority. Action(s) from the ACL entry 
+     * residing in the higher priority table within the group will be performed.
+     *
+     * Across the ACL groups, one hit ACL entry from each group will be selected 
+     * and all non-conflicting actions from them will be performed.
```
lguohan: per our discussion, add following comments here. "For conflicting actions, the ACL entry resides in the highest table priority wins."

ramangupta: Line added

PR #112, opened 2015-06-29
Modified FDB callback API to inform multiple entries in a single callback

dgsudharsan:


lguohan:
it seems like your pull request contains many unrelated commits. Not sure what exactly happens here, maybe because you need to generate the pull request against the latest master HEAD?

lguohan:
merged, thanks

PR #113, opened 2015-06-29
Added an attribute to get the list of ports belonging to a VLAN

dgsudharsan:


sai/inc/saitypes.h
```diff
@@ -179,6 +179,19 @@ typedef struct _sai_vlan_list_t {
 
 } sai_vlan_list_t;
 
+/*
```
lguohan: need to add a forward declaration here for sal_vlan_port_t as this one is defined in saivlan.h

sai/inc/saitypes.h
```diff
@@ -179,6 +179,19 @@ typedef struct _sai_vlan_list_t {
...
+ * Defines a vlan port list datastructure
+ */
+typedef struct _sai_vlan_port_list_t {
+
+    /* Number of ports in a VLAN*/
+    uint32_t vlan_port_count;
```
lguohan: rename this to count following other x_list structure

lguohan:
merged, thanks.

PR #115, opened 2015-06-29
Updated spec.md with default VLAN behavior

dgsudharsan:
Updated spec.md with description of behavior of default VLAN


sai/doc/spec.md
```diff
@@ -247,6 +247,38 @@ Provides FDB entries manipulation as well as aging/learning notifications.
...
+Initial VLAN Table
+VLAN ID	Ports
+1*	     1, 2
+10	
+
+
+Add port 1 to VLAN 10
+VLAN ID	Ports
+1*	     2
+10	     1
```
lguohan: in this example, if you add port 1 to vlan 10 using sai_add_ports_to_vlan_fn, then port 1 should belong to both vlan 1 and vlan 10. There is a port attribute SAI_PORT_ATTR_DEFAULT_VLAN, if you set that to 10, then its default vlan will become 10. Is my understanding correct here?

sai/doc/spec.md
```diff
@@ -247,6 +247,38 @@ Provides FDB entries manipulation as well as aging/learning notifications.
...
+
+Change Default VLAN to 2
+VLAN ID	Ports
+2*	     2
+10	     1
```
lguohan: I assume you are using SAI_PORT_ATTR_DEFAULT_VLAN to do it?

sai/doc/spec.md
dgsudharsan: Hi Guohan,
A port will be a member of default VLAN only if it is not a member of any VLAN(both tagged and untagged mode). If a port is added to some VLAN as a member in any tagging mode, then it is removed from default VLAN. So if port 1 is added to VLAN 10 then it is removed from VLAN 1(default VLAN). SAI_PORT_ATTR_DEFAULT_VLAN is used to set the PVID of the incoming untagged packets(native VLAN on a port).

lguohan: in the SAI a port can join multiple vlans (multiple vlan in tagged mode and one vlan in untagged mode). My understanding is that this is the default port mode.

dgsudharsan: Hi Guohan,
Yes in case of a port being in hybrid mode it can(Meaning that a port can exist as tagged as well as untagged in different VLANS). However my assumption is the port is in default mode is not hybrid. If not in hybrid mode, a port can be either in tagged or untagged state (not both together). Hence if a port is added as even tagged to a VLAN default VLAN is removed. This behavior can be controlled by having a port vlan mode attribute in port which might say the port is in hybrid or normal mode.

lguohan: Do we need to classify a port into hybrid/normal mode at SAI level or can we leave such classification outside of the SAI?

dgsudharsan: I agree that this classification is unnecessary at the SAI level if not for automating the default VLAN ports mechanism. The drawback without this port mode would be that adapter host or any layers above SAI cannot use the default VLAN in hybrid mode. A work around would be to allow the adapter host to modify default VLAN using add and delete port calls so that they can add default VLAN in required at hybrid ports alone but this might not look clean than having a port mode and having SAI to do the entire automation of default VLAN port handling.

lguohan:
merged, thanks.

PR #116, opened 2015-06-29
Update spec.md for Next Hop functionality

SathishKumarTR-Dell:
Updated the documentation under Next Hop functionality for not mandating the user to create the IP neighbor before creating the IP next hop. This was discussed in the opencompute-networking mailing list under the title "SAI Next Hop functionality description".


sai/doc/spec.md
```diff
@@ -271,7 +271,7 @@ Provides IP routing table (FIB) management functions. When an IP packet is recei
...
+Provides next hop objects such as IP next hop. The IP next hop is identified by a next hop ID which can be used in sai_route as well as other matching rules to forward traffic to an IP prefix to it. Intuitively, IP neighbor represents a device without any packet forwarding capability so that only traffic destined to its IP are forwarded to it, whereas IP next hop represents a device that can also forward other traffic. IP next hop is associated with a specific router interface, and is identified with an IP address. SAI associates IP next hop and the IP neighbor using the router interface ID and the IP address. User can create the IP next hop object and the IP neighbor entry independently. If user creates IP next hop object that do not have the corresponding IP neighbor entry, the next hop can be typically programmed to trap packets to the control plane for resolving the IP neighbor entry. When IP neighbor entry is created, packets will be forwarded to the IP next hop using the IP neighbor attributes.
```
lguohan: "If user creates IP next hop object that do not have the corresponding IP neighbor entry, the next hop can be typically programmed to trap packets to the control plane for resolving the IP neighbor entry."

When a sai_route points to an unresolved IP next hop, the packet behavior must be specified, can either be trap or drop. Where do we specify the option to define the packet behavior. Right now, there is no attribute in the sai_next_hop_attr_t. Should we define an attribute to define the packet action for unsolved case?

SathishKumarTR-Dell:
Perhaps we can have a global switch attribute (read-write) for this forwarding action. Default can be to trap to CPU. Having it in the sai_next-hop_attr_t will be overlapping with the current IP neighbor attribute SAI_NEIGHBOR_ATTR_PACKET_ACTION. Also, user need not update the attribute after the next hop gets resolved. Is this fine?

PR #119, opened 2015-07-02
ACL substages

kasubra:
3 ingress substages + placeholder for egress substages

### DIFF
--- a/sai/inc/saiacl.h
+++ b/sai/inc/saiacl.h
@@ -40,6 +40,18 @@ typedef enum _sai_acl_stage_t
     /* Egress Stage */
     SAI_ACL_STAGE_EGRESS,
 
+	/* Ingress Sub Stages */
+	/* Ingress Sub Stage Port ACL */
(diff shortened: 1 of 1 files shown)


sai/inc/saiacl.h
```diff
@@ -40,6 +40,18 @@ typedef enum _sai_acl_stage_t
     /* Egress Stage */
     SAI_ACL_STAGE_EGRESS,
 
+	/* Ingress Sub Stages */
+	/* Ingress Sub Stage Port ACL */
```
lguohan: can you add description for each substage?

lguohan:
".. on all fields of the native packet before any editing or transformations .. " This is not very accurate as if the packets go through L2 forwarding, it wouldn't be edited or transformed. But the asic will generate some meta data at this stage. Therefore, I think positioning this stage in relative with other component is more intuitive and accurate. And I think it's better to phrase them for all the substages here.

PR #121, opened 2015-07-07
Fix overflow bug in stub_object_id_t

itaibaz:
Object was of size 12 bytes instead of 8 bytes (as original
sai_object_id_t)


lguohan:
merged, thanks

PR #123, opened 2015-07-08
SAI Acl table group support 

ramangupta:


sai/inc/saiacl.h
```diff
@@ -136,6 +136,40 @@ typedef enum _sai_acl_table_attr_t
      * allocated and able to do ACL CAM Carving accurately */
     SAI_ACL_TABLE_ATTR_SIZE,
 
+    /* Table Group Id [sai_uint32_t], CREATE_ONLY
+     *
+     * By default, if the attribute is not present during table creation, SAI 
+     * will internally allocate a unique group id which can be used to group
+     * tables which are created later. User would perform get operation for the 
+     * already created ACL Table to fetch the group id created by SAI and use
+     * it for grouping the next table at the time of its creation.
```
lguohan: Group Id will be automatically deleted when all ACL tables that use this group id are deleted.

ramangupta: Will include this point in the comments

lguohan:
change the id type to sai_object_id_t and merged, thanks.

PR #125, opened 2015-07-08
Add route attribute directly reachable

itaibaz:


sai/inc/sairoute.h
```diff
@@ -52,6 +52,11 @@ typedef enum _sai_route_attr_t
      * next hop group. */
     SAI_ROUTE_ATTR_NEXT_HOP_ID,
 
+    /* Is on-link route [bool]
+       ASICs that don't need this indication, can ignore this attribute
+       (default to false) */
+    SAI_ROUTE_ATTR_DIRECTLY_REACHABLE_ROUTE,
```
lguohan: suggest to change the following comments to make this clear.

"Is directly reachable route [bool]
Directly reachable routes are the IP subnets that are directly attached to the router."

suggest to remove "ASICs that don't need this indication, can ignore this attribute",
because once the definition for the attribute is clear, asic vendor should know what to do.
No need to specify it.

itaibaz: OK. I modified the pull request accordingly
Thanks

PR #126, opened 2015-07-12
New Actions added to support ECN and other packet actions

ramangupta:


sai/inc/saiswitch.h
```diff
@@ -68,11 +68,14 @@ typedef enum _sai_packet_action_t
...
+    /* Copy Packet to CPU and Drop (CPU + Drop) */
     SAI_PACKET_ACTION_TRAP,
 
-    /* Log (Trap + Forward) Packet */
-    SAI_PACKET_ACTION_LOG
+    /* Copy Packet to CPU */
+    SAI_PACKET_ACTION_LOG,
+
+    /* Copy Packet to CPU and Forward (CPU + Forward) */
+    SAI_PACKET_ACTION_LOG_FORWARD
```
lguohan: current modification change meaning of LOG, can we still keep LOG as the original meaning and just add a SAI_PACKET_ACTION_COPY to copy only action?

ramangupta: That would be fine.

sai/inc/saiswitch.h
```diff
@@ -68,10 +68,13 @@ typedef enum _sai_packet_action_t
     /* Forward Packet */
     SAI_PACKET_ACTION_FORWARD,
 
-    /* Trap Packet to CPU */
+    /* Copy Packet to CPU and Drop (CPU + Drop) */
     SAI_PACKET_ACTION_TRAP,
```
atitjain: trap and log is clear, what is copy.

ramangupta: Trap and Log explicitly implies that packet will be dropped or forwarded respectively along with being copied to CPU. This would override any conflicting actions picked from subsequent entry hits residing at a lower priority. 
Copy implies that the packet will only be sent to CPU. Packet could be either dropped or forwarded depending upon the lookup result in the pipeline.

ramangupta:
Closing this pull request as most of it is merged. Will take up packet action change in a new pull request.

PR #127, opened 2015-07-14
VLAN FDB Learn mode

dgsudharsan:
Introducing learn mode for VLAN as in port. For certain features like null VLAN where learning and forwarding are disabled there must be VLAN level attribute to control forwarding. Since the port learn modes can also apply to a VLAN it has been generalized and moved to saitypes.h


atitjain:
If learn mode is set to port and vlan both, what is the behavior. Which is applied/considered?

dgsudharsan:
Hi Atit,
Current instead of learn mode its learn disable on a VLAN. Below is the comment from file saivlan.h
 
   /** To disable learning on a VLAN. [bool] (CREATE_AND_SET)
      * (default set to false)
      * This should override port learn settings. If this is set to true on a vlan,
      * then the source mac learning is disabled for this vlan on a member port even
      * if learn is enable on the port(based on port learn attribute)
      */
    SAI_VLAN_ATTR_LEARN_DISABLE,

PR #128, opened 2015-07-15
MOdified int to uint32_t

AshokDaparthi:
WRED thresholds are mandatory to enable WRED by color based. This will make SAI clean.  
Disable one of the color WRED means internally SAI needs to take care of not enable on particular color by setting the min and max thresholds to max buffer limit value if NPU does not support separate knob to do this.


lguohan:
merged, thanks

PR #132, opened 2015-07-16
Enable/disable learning on a VLAN

dgsudharsan:
To enable/disable learning on a VLAN. For certain features like RSPAN it is required that learning can be disabled on a VLAN and thus SAI must provide an attribute to disable/enable learning on a VLAN. This should override port learn settings if both port learn and vlan learn settings are enabled on a port.


sai/inc/saivlan.h
```diff
@@ -80,6 +80,11 @@ typedef enum _sai_vlan_attr_t
      * (default to default stp instance id)*/
     SAI_VLAN_ATTR_STP_INSTANCE,
 
+    /* To disable learning on a VLAN. [bool]
+     * (default set to false)
+     */
+    SAI_VLAN_ATTR_LEARN_DISABLE,
+
```
lguohan: can you add "This should override port learn settings. If this is set to true on a vlan, then the source mac learning is disabled for this vlan for a port even learn is enable on the port" to clarify its relation with the port-based mac learning?

dgsudharsan: Done

lguohan:
merged, thanks.

PR #133, opened 2015-07-17
Renamed port default vlan id to port vlan id

dgsudharsan:
Renamed port attribute default vlan to port attribute port vlan id(PVID) to avoid confusing with switch attribute default VLAN.


lguohan:
merged, thanks.

PR #135, opened 2015-07-20
Qos patch

AshokDaparthi:
- Added the SAI_PORT_ATTR_QOS_NUMBER_OF_QUEUES attribute.
- Removed the SAI_QUEUE_DROP_TYPE attributes, Like other objects wred ID other than the SAI_NULL_OBJECT_ID means WRED enabled of queue. Default drop type is TAIL drop. there is no need for separate knob to control WRED.
- Modified the descriptions for WRED thresholds.


lguohan:
merged, thanks.

PR #136, opened 2015-07-21
New Packet Action proposed

ramangupta:
Since packet action would simply replace any existing packet action, all combinations of packet actions which are non-conflicting should be supported. A new packet action is introduced which would simply copy the packet to CPU without adding any further action such as drop and forward. This would provide flexibility to user to decide on which action to use. 

Different flavors of packet action to punt a packet to CPU are:
COPY - Just send copy of packet to CPU. Drop/Forward will be based on the lookup results in the pipeline. 
TRAP - Copy sent to CPU and explicitly adding drop. Drop will supersede and will override the lookup result.   
LOG - Copy sent to CPU and explicitly adding forward. Forward will supersede and will override the lookup result.


dplore:
Is the intent in SAI that only one packet action is allowed per packet?

If a packet matches and ACL and a packet_action, what takes precedence?

Is there a reason sai_packet_action cannot instead be defined as an ACL?  (where the CPU is represented as a destination port)  If so, using only ACL's would be simpler than having packet_action and ACL's.  Is there a scenario that requires a "packet_action" that cannot be served by ACL?

If one can define the packet_actions as ACL’s, one can represent COPY, TRAP, LOG as ACL's and we don't need more packet_actions?

ramangupta:
1) There could be 2 possibilities how the usage of packet action can be defined. Either we just have basic packet actions such as drop, forward and copy and can be used in conjunction to perform combination of actions such as trap or log OR a predefined set of possible packet actions are provided which are non-conflicting and can be easily used based on the need. In first approach, packet actions need to be validated against conflicts. 
2) Though the behavior is based on vendor-specific pipeline, but if there is a conflict of packet action between ACL and any other stage (say L3), ACL should take precedence as a standard behavior. This is how ACL can control/limit traffic. I think the order of precedence on packet actions should be [Policy-based Routing > ACL > Conventional Routing]  
3) Packet Actions are generalized so that the usage can be spread across different modules.

sai/inc/saiswitch.h
```diff
@@ -62,17 +62,23 @@ typedef enum _sai_switch_oper_status_t
...
 
-    /* Forward Packet */
+    /* Forward Packet in data plane */
     SAI_PACKET_ACTION_FORWARD,
 
-    /* Trap Packet to CPU */
+    /* Copy Packet to CPU. Forwarding decision will be based on the lookup results in the NPU's Pipeline */
+    SAI_PACKET_ACTION_COPY,
+
+    /* Copy Packet to CPU and Drop the packet in data plane (CPU + Drop) */
```
lguohan: (CPU + Drop). This is a combination of sai packet action COPY and DROP.

sai/inc/saiswitch.h
```diff
@@ -62,17 +62,23 @@ typedef enum _sai_switch_oper_status_t
...
     SAI_PACKET_ACTION_TRAP,
 
-    /* Log (Trap + Forward) Packet */
-    SAI_PACKET_ACTION_LOG
+    /* Copy Packet to CPU and Forward the packet in data plane (CPU + Forward) */
```
lguohan: (CPU + Forward): This is a combination of sai packet action (COPY + FORWARD)

sai/inc/saiswitch.h
```diff
@@ -62,17 +62,23 @@ typedef enum _sai_switch_oper_status_t
 */
 typedef enum _sai_packet_action_t
 {
-    /* Drop Packet */
+    /* Drop Packet in data plane */
     SAI_PACKET_ACTION_DROP,
 
-    /* Forward Packet */
+    /* Forward Packet in data plane */
```
lguohan: should we add "in case the packet is marked as drop, this will cancel drop action."?

sai/inc/saiswitch.h
```diff
@@ -62,17 +62,23 @@ typedef enum _sai_switch_oper_status_t
...
+    /* Copy Packet to CPU and Drop the packet in data plane (CPU + Drop) */
     SAI_PACKET_ACTION_TRAP,
 
-    /* Log (Trap + Forward) Packet */
-    SAI_PACKET_ACTION_LOG
+    /* Copy Packet to CPU and Forward the packet in data plane (CPU + Forward) */
+    SAI_PACKET_ACTION_LOG,
+    
+    /* Packet not sent to CPU and Drop in data plane */
+    SAI_PACKET_ACTION_DENY
```
lguohan: should we say "Cancel copy the packet to cpu and drop the packet in data plane"? 
It looks like we are missing the SAI_PACKET_ACTION_COPY_CANCEL action? 

The four basic types are DROP, FORWARD, COPY and COPY_CANCEL.
This first two deal with the packet action in the data plane, the latter two defines the packet action in the cpu path.

Then we can four combinations:
TRAP = COPY + DROP
LOG = COPY + FORWARD
DENY = COPY_CANCEL + DROP
??? = COPY_CANCEL + FORWARD

ramangupta: We can call COPY_CANCEL + FORWARD = TRANSIT. I have made the change in the pull request.

lguohan:
merged

PR #137, opened 2015-07-21
Add L2 Netdev channel. Remove Sample packet user defined trap

itaibaz:
Sample packet user defined trap already exists as regular trap


sai/inc/saihostintf.h
```diff
@@ -252,9 +252,12 @@ typedef enum _sai_hostif_trap_channel_t
     /* receive packets via callback */
     SAI_HOSTIF_TRAP_CHANNEL_CB,
 
-    /* receive packets via OS net device */
+    /* receive packets via OS L3 net device */
     SAI_HOSTIF_TRAP_CHANNEL_NETDEV,
 
+    /* receive packets via OS L2 net device */
+    SAI_HOSTIF_TRAP_CHANNEL_L2_NETDEV,
```
lguohan: We need more detailed description for introducing the L2 NETDEV into the model here.

sai/inc/saihostintf.h
```diff
@@ -338,11 +341,8 @@ typedef sai_status_t(*sai_get_hostif_trap_attribute_fn)(
  */
 typedef enum _sai_hostif_user_defined_trap_id_t
 {
-    /* Samplepacket traps  */
-    SAI_HOSTIF_USER_DEFINED_TRAP_ID_SAMPLEPACKET = 0x00000001,
-
```
lguohan: can you explain why this trap id is removed?

itaibaz: We have both regular trap for sample packet
    /* default action is trap */
    SAI_HOSTIF_TRAP_ID_SAMPLEPACKET = 0x00000000b,

And user defined trap for sample packet

    /* Samplepacket traps  */

    SAI_HOSTIF_USER_DEFINED_TRAP_ID_SAMPLEPACKET = 0x00000001,

So I think we should be fine with removing the user defined flavor

PR #138, opened 2015-07-22
add PFC-related qos maps

lguohan:
SAI_QOS_MAP_TC_TO_PRIORITY_GROUP is used to map incoming traffic to
priority group based on its TC.

SAI_QOS_MAP_PRIORITY_GROUP_TO_PFC_PRIORITY is used to set the priority
for pause frame for congested priority group.

SAI_QOS_MAP_PFC_PRIORITY_TO_TC is used to stop transmit corresponding
TC based on the priority in the received pause frame.


lguohan:
merged

PR #140, opened 2015-07-24
add user/npu meta for fdb, neighbor, port, route, vlan

lguohan:


sai/inc/saiacl.h
```diff
@@ -269,8 +269,36 @@ typedef enum _sai_acl_table_attr_t
...
+    /* Meta Data carried from previous ACL Stage */
+    SAI_ACL_TABLE_ATTR_FIELD_ACL_LOOKUP_USER_META,
+
+    /* NPU Based Meta Data [bool] */
+
+    /* DST MAC address match in FDB */
+    SAI_ACL_TABLE_ATTR_FIELD_FDB_DST_NPU_META_HIT,
+
+    /* DST IP address match in L3 Table */
+    SAI_ACL_TABLE_ATTR_FIELD_L3_DST_NPU_META_HIT,
```
lguohan: L3 Table: is it route table or neighbor table, or both?

sai/inc/saiacl.h
```diff
@@ -403,8 +431,48 @@ typedef enum _sai_acl_entry_attr_t
...
+
+    /* Vlan User Meta Data
+     * Value must be in the range defined in
+     * SAI_SWITCH_ATTR_VLAN_USER_META_DATA_RANGE */
+    SAI_ACL_ENTRY_ATTR_FIELD_VLAN_USER_META,
+
+    /* Meta Data carried from previous ACL Stage
+     * Value must be in the range defined in
+     * SAI_SWITCH_ATTR_ACL_LOOKUP_USER_META_DATA_RANGE */
+    SAI_ACL_ENTRY_ATTR_FIELD_ACL_LOOKUP_USER_META,
```
lguohan: here we assume all acl generate and consume one acl user meta. What does the lookup mean here?

sai/inc/saiacl.h
```diff
@@ -269,8 +269,36 @@ typedef enum _sai_acl_table_attr_t
...
+    SAI_ACL_TABLE_ATTR_FIELD_NEIGHBOR_DST_USER_META,
+
+    /* Port User Meta Data */
+    SAI_ACL_TABLE_ATTR_FIELD_PORT_USER_META,
+
+    /* Vlan User Meta Data */
+    SAI_ACL_TABLE_ATTR_FIELD_VLAN_USER_META,
+
+    /* Meta Data carried from previous ACL Stage */
+    SAI_ACL_TABLE_ATTR_FIELD_ACL_LOOKUP_USER_META,
```
lguohan: what does the lookup mean here?

ramangupta: Implies that the meta data is obtained from the lookup of some previous ACL stage which needs to be qualified in the present stage. This is the reason we have a new ACL entry action SAI_ACL_ENTRY_ATTR_ACTION_SET_ACL_METADATA to set the metadata.

sai/inc/saiacl.h
ramangupta: This is NPU meta data which is just a hit bit and cannot be differentiated between Route and Neighbor. So it will be applicable for both.

sai/inc/saiacl.h
ramangupta: Lookup simply signifies that this meta data was generated based on the lookup of a former ACL stage

sai/inc/saiacl.h
lguohan: This assumes all ACL stages share a same meta data. At any stage, when an ACL entry set the meta data, it overrides the meta data carried from previous stages. I do not know how universal this model to be, but it does seem like a simple model to carry acl meta data. However, I think it's better to document the above assumption in the comments to make it clear. Agree?

BTW, the lookup does not seem to convey more information here. Can we remove it to align with other naming convention?

ramangupta: I am fine with adding comment to describe the ACL stage meta data behavior. Lookup can also be removed from ACL meta data. I guess wouldn't be able to edit the pull request as it was not raised by me.

lguohan:
merged, thanks.

PR #142, opened 2015-07-30
Removed unwanted include files and Corrected spell mistakes.

AshokDaparthi:
- Removed sai.h from all sai qos objects file.
- Corrected some typos.
- Added one maps type missed which is there previously.
- Drop type attribute is removed so removed corresponding enum.


lguohan:
merged, thanks.

PR #143, opened 2015-07-31
Unit Test: basic_router v1.1

xdwu:
<EOM>

### DIFF
--- a/sai/Makefile
+++ b/sai/Makefile
@@ -1,0 +1,68 @@
+#	 Copyright (c) 2015 Microsoft Open Technologies, Inc.
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may 
+#    not use this file except in compliance with the License. You may obtain 
+#    a copy of the License at http://www.apache.org/licenses/LICENSE-2.0
+#
+#    THIS CODE IS PROVIDED ON AN  *AS IS* BASIS, WITHOUT WARRANTIES OR 
+#    CONDITIONS OF ANY KIND, EITHER EXPRESS OR IMPLIED, INCLUDING WITHOUT 
+#    LIMITATION ANY IMPLIED WARRANTIES OR CONDITIONS OF TITLE, FITNESS 
+#    FOR A PARTICULAR PURPOSE, MERCHANTABLITY OR NON-INFRINGEMENT.
+#
+#    See the Apache Version 2.0 License for specific language governing 
+#    permissions and limitations under the License. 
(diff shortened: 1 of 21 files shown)


itaibaz:
Hi,
I have reviewed the host interface related code in the unit test
I notice you set action=trap for :
TTL error
Arp request
Arp response
LLDP
And in addition you create host interface of type netdev over ports

As far as I understand, the default trap channel is callback
So for real traffic over the port netdev, I think you should also set per above trap, the trap channel to be netdev
attr.id = SAI_HOSTIF_TRAP_ATTR_TRAP_CHANNEL;
attr.value.s32 = SAI_HOSTIF_TRAP_CHANNEL_NETDEV;

In addition, no need to set the trap action for TTL error to trap, as by default it’s trap. No harm in that, though it’s redundant.

Lastly, for user defined traps, if you want them to work over the port netdev, setting the channel to netdev is also needed, as far as I understand
For Mellanox ASIC, for example, to have real traffic, you will need to set the channel for :
SAI_HOSTIF_USER_DEFINED_TRAP_ID_ROUTER_MIN
SAI_HOSTIF_USER_DEFINED_TRAP_ID_NEIGH_MIN
And when you add ACLs, also for
SAI_HOSTIF_USER_DEFINED_TRAP_ID_ACL_MIN


Other than that, it’s great help having this application as reference, and I plan to run it in our regression.

Few possible additions in the future :

1)      Currently seems like only positive flows are tested.

We might want to add also negative flows with illegal config.

2)      Having real traffic. Defining a common topology – for example, connect host A to the first port, host B to the 2nd port, etc.

This will add a lot of complexity, but can give good benefits

Thanks

xdwu:
Hello, Itai 

Thanks for your comments.
* You are right, we do need to add this configuration. I will address this later. 
attr.id = SAI_HOSTIF_TRAP_ATTR_TRAP_CHANNEL;
attr.value.s32 = SAI_HOSTIF_TRAP_CHANNEL_NETDEV;

* set the trap action for TTL error by default by MLNX
Different vendors might have different opinions on what should be the default.
Here, based on SAI specification, we assume trap action is not set 
for TTL error by default. 

* We do not addressed user defined traps in this version. We will take care of 
these in the future.

I totally agree with you on what we should do in the future.
1)  adding more negative tests
2)  Injecting real traffic in the test cases.  

Thanks
xdwu

lguohan:
merged, thanks.

PR #145, opened 2015-08-04
Added ACL counter types to get green/yellow/red packets/bytes.

AshokDaparthi:


sai/inc/saiacl.h
```diff
@@ -609,6 +609,24 @@ typedef enum _sai_acl_counter_attr_t
     /* get/set byte count [uint64_t] */
     SAI_ACL_COUNTER_ATTR_BYTES
 
+    /* get/set green packet count [uint64_t] */
+    SAI_ACL_COUNTER_ATTR_GREEN_PACKETS,
```
lguohan: 1. how does this differ from sai_policer_stat_counter_t? 
2. do we need to have the enable attribute to enable/disable these packet counts?

AnbalaganNatchimuthu: 1. Some of the NPUs do not support policer directly. We need to apply the policer to the ACL rule to acheive metering, These counters are needed for such cases. When acl rule is created and policer is applied as an action to acl rule these counters can be enabled to get policer stats. In NPUs which support policer stats directly the sai_policer_stat_counter_t attributes can be used.

2. We will add enable/disable for the policer counters. The attributes added are for get/set.

sai/inc/saiport.h
```diff
@@ -387,9 +387,6 @@ typedef enum _sai_port_attr_t
     * Default no map */
     SAI_PORT_ATTR_QOS_TC_AND_COLOR_TO_DSCP_MAP,
 
-   /* sai_qos_drop_type_t, Default (TAIL DROP)*/
-    SAI_PORT_ATTR_QOS_DROP_TYPE,
-
    /* Attach WRED to port [sai_object_id_t]
      (mandatory when SAI_PORT_ATTR_QOS_DROP_TYPE =  SAI_QOS_DROP_TYPE_WRED) */
```
lguohan: as DROP_TYPE is remove, the comment here also needs to be removed.

PR #146, opened 2015-08-07
Add attribute to query supported speeds for a port, get/set full duplex and autoneg configuration

Rajasekhar-Pulluru:


sai/inc/saiport.h
```diff
@@ -222,10 +247,19 @@ typedef enum _sai_port_attr_t
     /* List of Scheduler groups for the port[sai_object_list_t] */
     SAI_PORT_ATTR_QOS_SCHEDULER_GROUP_LIST,
 
+    /* Query list of supported port speed [sai_s32_list_t of sai_port_speed_t] */
+    SAI_PORT_ATTR_SUPPORTED_SPEED,
+
     /* READ-WRITE */
     /* Speed in Mbps [uint32_t] */
     SAI_PORT_ATTR_SPEED,
```
lguohan: do we need to modify this attribute value to be sai_port_speed_t defined above?

sai/inc/saiport.h
```diff
@@ -222,10 +247,19 @@ typedef enum _sai_port_attr_t
     /* List of Scheduler groups for the port[sai_object_list_t] */
     SAI_PORT_ATTR_QOS_SCHEDULER_GROUP_LIST,
 
+    /* Query list of supported port speed [sai_s32_list_t of sai_port_speed_t] */
+    SAI_PORT_ATTR_SUPPORTED_SPEED,
```
lguohan: Similiar to SAI_PORT_ATTR_SPEED, can we make the attribute value as sai_u32_list_t and list all the supported speed in Mbps. It's more flexible and we do not need to define sai_port_speed_t. For example, it the port support 25G or 50G, we do not need to modify sai_port_speed_t.

sai/inc/saiport.h
```diff
@@ -222,10 +247,19 @@ typedef enum _sai_port_attr_t
...
+
     /* READ-WRITE */
     /* Speed in Mbps [uint32_t] */
     SAI_PORT_ATTR_SPEED,
 
+    /* Full Duplex setting [bool] (default to TRUE) */
+    SAI_PORT_ATTR_FULL_DUPLEX_MODE,
+
+    /* AutoNegotiation configuration [bool] (Default set to FALSE) */
+    SAI_PORT_ATTR_AUTONEG_MODE,
```
lguohan: why not make this default as TRUE?

sai/inc/saiswitch.h
```diff
@@ -132,6 +132,9 @@ typedef enum _sai_switch_attr_t
     /* Get the port list [sai_object_list_t] */
     SAI_SWITCH_ATTR_PORT_LIST,
 
+    /* Get the Max MTU Supported by the switch [uint32_t] */
```
lguohan: can we add the "Max MTU in bytes"

lguohan:
merged, thanks. BTW, please generate the pull request based on the lastest master head, it then be easier to merge the pull request into the master branch.

PR #147, opened 2015-08-12
Add doxygen support

itaibaz:
1) Modify H files comments to doxygen format
2) Add doxyfile - doxygen config used to create HTMLs
3) Add HTML dir


lguohan:
merged, thanks. The Makefile to generate doxygen files are also added.

PR #149, opened 2015-08-24
Lag Port mode changes

dgsudharsan:
To enable or disable distribution and collection of traffic for a port as part of LAG

### DIFF
--- a/sai/inc/sailag.h
+++ b/sai/inc/sailag.h
@@ -35,40 +35,10 @@ typedef enum _sai_lag_attr_t {
 
     /** READ_WRITE */
 
-    /** SAI port list [sai_object_list_t] (CREATE_AND_SET) */
-    SAI_LAG_ATTR_PORT_LIST,
-
-    /** SAI port mode list [sai_lag_port_mode_list_t] (CREATE_AND_SET) */
-    SAI_LAG_ATTR_PORT_MODE_LIST
+    /** Custom range base value */
+    SAI_LAG_ATTR_CUSTOM_RANGE_BASE  = 0x10000000
 
 } sai_lag_attr_t;
-
-/**
- *  @brief List of LAG Port member modes
- */
-typedef enum _sai_lag_port_member_mode_t {
-
-    /** Port forwards and receives traffic as part of LAG. */
-    SAI_LAG_PORT_MEMBER_MODE_FORWARD,
-
-    /** Disable traffic distribution to this port as part of LAG. */
-    SAI_LAG_PORT_MEMBER_MODE_EGRESS_DISABLE,
-
-    /** Disable traffic collection from this port as part of LAG. */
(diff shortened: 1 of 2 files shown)


sai/inc/sailag.h
```diff
@@ -116,6 +131,38 @@ typedef sai_status_t (*sai_get_lag_attribute_fn)(
...
+    \param[out] lag_port_mode_list List of lag port modes obtained for the ports in port_list
+    \return Success: SAI_STATUS_SUCCESS
+            Failure: Failure status code on error
+*/
+typedef sai_status_t (*sai_lag_port_mode_get_fn)(
+    _In_ sai_object_id_t lag_id,
+    _In_ uint32_t count,
+    _In_ const sai_object_id_t *port_list,
+    _Out_ sai_lag_port_mode_t *lag_port_mode_list
+    );
```
lguohan: can we have define a attribute value structure to include the lag port mode, so that we can reuse get/set_lag_attribute to control the traffic distribution/collection mode for the port.

sai/inc/sailag.h
```diff
@@ -37,8 +37,39 @@ typedef enum _sai_lag_attr_t {
...
+
+    /** SAI port mode list [sai_lag_port_mode_list_t] (CREATE_AND_SET) */
+    SAI_LAG_ATTR_PORT_MODE_LIST
+
 } sai_lag_attr_t;
 
+/**
+ *  @brief List of LAG Port member modes
+ */
+typedef enum _sai_lag_port_member_mode_t {
```
atitjain: What is the usecase here. How does it work with LACP etc.

dgsudharsan: In order for LACP to add port to a LAG without traffic loss, it is desirable that data flow occurs after both the ends are programmed by the protocol. To achieve it provisions are required to disable traffic distribution and collection initially during protocol establishment. After the programming is done at both the ends by protocol first collection is enabled at one end and then distribution at the peer (also repeating it in the same order in the other direction).

sai/inc/sailag.h
```diff
@@ -64,54 +65,105 @@ typedef sai_status_t(*sai_remove_lag_fn)(
...
+ */
+typedef enum _sai_lag_member_attr_t {
+
+    /** READ_WRITE */
+
+    /* lag ID [sai_object_id_t] (MANDATORY_ON_CREATE|CREATE_ONLY) */
+    SAI_LAG_MEMBER_ATTR_LAG_ID,
+
+    /* logical port ID [sai_object_id_t] (MANDATORY_ON_CREATE|CREATE_ONLY) */
+    SAI_LAG_MEMBER_ATTR_PORT_ID,
```
lguohan: is it ok to make this attribute CREATE_AND_SET. The use case is to replace a lag member.

lguohan:
merged, thanks.

PR #150, opened 2015-08-24
AutoNeg is set to FALSE by default. 

Rajasekhar-Pulluru:
…t set to

FALSE for port with speed >= 10G

 Changes to be committed:
	modified:   inc/saiport.h


lguohan:
thanks

PR #153, opened 2015-09-02
sai switch attribute enhancements

Rajasekhar-Pulluru:
Added new argument to sai switch initialize function to pass platform specific configuration file.
Added new attributes.
Modified uint32_t to sai_uint32_t

 Changes to be committed:
	modified:   inc/saiswitch.h
	modified:   inc/saitypes.h


itaibaz:
Can we add standard K/V table ?
I guess it should hold all the new attributes in this commit
A const array of strings should be fine
Thanks

arunsubash:
Yes Tushar "Bytes" should apply only for MTU and for other tables it will be removed.

arunsubash:
Yes Itai we will add a constant array of strings in saitypes.h that lists the attributes that can be set at init time

sai/inc/saiswitch.h
```diff
@@ -428,6 +499,34 @@ typedef enum _sai_switch_attr_t
...
+ * @def SAI_SWITCH_ATTR_MAX_KEY_STRING_LEN
+ * Maximum length of switch attribute key string that can be set using key=value 
+ */
+#define SAI_SWITCH_ATTR_MAX_KEY_STRING_LEN    64
+
+/**
+ * @def SAI_SWITCH_ATTR_MAX_KEY_COUNT
+ * Maximum count of switch attribute keys 
+ */
+#define SAI_SWITCH_ATTR_MAX_KEY_COUNT         11
```
arunsubash: Add a comment to indicate that this needs to be incremented every time we add a new define below

Rajasekhar-Pulluru:
Hi Itai, As discussed offline, since SAI could also be used as a library and we can't define const array of strings in header file, defined new macros for switch attribute keys.

PR #154, opened 2015-09-06
Add SAI_ROUTE_ATTR_DIRECTLY_REACHABLE_EGRESS_RIF

itaibaz:
The router interface to which the subnet is attached

### DIFF
--- a/sai/inc/sairoute.h
+++ b/sai/inc/sairoute.h
@@ -67,6 +67,12 @@ typedef enum _sai_route_attr_t
      * Value Range SAI_SWITCH_ATTR_ROUTE_USER_META_DATA_RANGE */
     SAI_ROUTE_ATTR_META_DATA,
 
+    /** The router interface to which the subnet is attached [sai_object_id_t] (CREATE_AND_SET)
+    * Valid only when SAI_ROUTE_ATTR_DIRECTLY_REACHABLE_ROUTE == true
+    *   (MANDATORY_ON_CREATE when SAI_ROUTE_ATTR_DIRECTLY_REACHABLE_ROUTE == true)
+    *   (default to SAI_NULL_OBJECT_ID) */
+    SAI_ROUTE_ATTR_DIRECTLY_REACHABLE_EGRESS_RIF,
+
     /** Custom range base value */
     SAI_ROUTE_ATTR_CUSTOM_RANGE_BASE  = 0x10000000
 


sai/inc/sairoute.h
```diff
@@ -67,6 +67,12 @@ typedef enum _sai_route_attr_t
      * Value Range SAI_SWITCH_ATTR_ROUTE_USER_META_DATA_RANGE */
     SAI_ROUTE_ATTR_META_DATA,
 
+    /** The router interface to which the subnet is attached [sai_object_id_t] (CREATE_AND_SET)
+    * Valid only when SAI_ROUTE_ATTR_DIRECTLY_REACHABLE_ROUTE == true
+    *   (MANDATORY_ON_CREATE when SAI_ROUTE_ATTR_DIRECTLY_REACHABLE_ROUTE == true)
+    *   (default to SAI_NULL_OBJECT_ID) */
+    SAI_ROUTE_ATTR_DIRECTLY_REACHABLE_EGRESS_RIF,
+
```
SathishKumarTR-Dell: Can this attribute be set along with the SAI_ROUTE_ATTR_NEXT_HOP_ID? 
Route pointing to a point-to-point IP tunnel will be a directly reachable route on the overlay network - for which, this new attribute can be set to the overlay tunnel RIF and the route can be still set with the nexthop id for the transport network?

itaibaz: Hi
At the moment, directly reachable=true + egress rif value, is mutually exclusive with giving a next hop ID
Matty is working on a proposal for tunneling, which he will present in the coming weeks for the community. We can consider then the specific changes needed to support tunneling, after we evaluate the model.

Thanks

lguohan: For directly reachable route, can we put the egress rif Id into the SAI_ROUTE_ATTR_NEXT_HOP_ID? SAI can look at the object type of it, if it is NEXT_HOP, then SAI uses the (IP, rif) in the next hop to lookup the neighbor table. If it is a ROUTER_INTERFACE, then, SAI uses the (dst IP of the packet, rif) to lookup the neighbor table.

itaibaz: Yes, it’s a good idea, as the modes are mutually exclusive it’s easier to use one attribute for all modes, and distinguish between them according to the object type – either next hop, or next hop group, or router interface
I will modify the pull request accordingly

PR #157, opened 2015-09-07
Remove switch default trap channel, fd. Make switch default trap grou…

itaibaz:
…p RO

Remove switch default trap channel and FD, that can be managed by host
trap channel and FD attributes
Modify the host trap channel and FD defaults
Modify switch default trap group to be read only. The object handle is
RO. The group attributes are RW, for example modifying the default group
policer and queue


lguohan:
thanks

PR #158, opened 2015-09-08
Port buffer profile changes

dgsudharsan:
Added missing buffer profile attribute for port.
Modified SAI_PORT_ATTR_PG_NUM to SAI_PORT_ATTR_NUM_PG and also changed comments for better clarity.


tushar-ty:
It seems that the definition of sai_qos_drop_type_t is still missing in the headers.

PR #159, opened 2015-09-11
Egress block port list

SathishKumarTR-Dell:
Added attribute to set an egress block list on a port.
Use-case is to achieve port isolation by preventing an ingress port from sending packets to a set of egress ports.


lguohan:
thanks

atitjain:
Hi Sathish,

What use case we are looking @ here.

SathishKumarTR-Dell:
This can be used for port security like blocking a specific port from forwarding packets to certain other port(s). Can also be used to block the flooded packets in a network topology having redundant links like the Multichassis LAG.

atitjain:
Just a thought, in the more networks moving to virtualized shouldn't we be thinking for port,vlan option rather than the port alone. Also for LAG on the egress how do we decide @ ingress port.

PR #160, opened 2015-09-15
ICMP Type Code field attribute added in SAI ACL

ramangupta:


sai/inc/saiacl.h
```diff
@@ -285,6 +285,9 @@ typedef enum _sai_acl_table_attr_t
     /** Class-of-Service (Traffic Class) */
     SAI_ACL_TABLE_ATTR_FIELD_TC,
 
+    /** ICMP Type and Code */
+    SAI_ACL_TABLE_ATTR_FIELD_ICMP_TYPE_CODE,
```
lguohan: a minor comment, can we separate the icmp type and icmp code into two matching fields in case an ASIC supports them separately.

ramangupta: Done

lguohan:
thanks

